PR #22, opened 2017-06-03
rrdesi frames and queues

sbailey:
@gdhungana please try out this rrdesi-frames branch of redrock with your ELG tests.  It fixes a queue flushing bug when there are hundreds of targets in a brick.  Previously I had smaller bricks succeeding but larger bricks were hanging.

Other changes:
  * switch to using `desispec.io.read_frame()` instead of `desispec.io.Brick()` to load brick files.  This gives more flexibility, e.g. by reading by EXTNAME instead of HDU number, and automatically taking care of endianness and float -> double conversions.
  * debugging printouts include the PID to make it easier to track which message go together when wrapped with MPI but still writing to a single log.
  * fixes bug where fitz was using multiprocessing even if ncpu==1

**Details for the record**

In the "fitz" step, redrock parallelizes over targets and uses a `multiprocessing.Queue` object to communicate results back to the parent process.  When there are hundreds of targets, the multiprocessing processes need to call `Queue.close()` to indicate that they are done putting things into the queue before exiting.  Otherwise it seems that the queue isn't completely flushed and the parent process hangs while waiting for results from the Queue.


gdhungana:
@sbailey Thanks for the update. I had my redshift fitting done for my current samples before this update came. In my case, each brick was O(50) targets. Although, there was hanging up the first time (using 40 nodes) after finishing ~140/147 bricks, later on all the bricks passed even with the master branch. I used 20 nodes second time. There will be more runs I'll need to do, but not very soon. Then I'll test on bricks with larger number of targets. Also thanks for fixing --ncpu =1 issue.

Issue #29, opened 2017-07-11
bricknames and targets


ngbusca:
l. 187 in external/desi.py asserts len(brickname)==len(targets)

My run on plate 4199 of boss2desi is crashing at this line since I have one target rejected due to

```
ERROR: Target 102696313 on brick-b-4199.fits has no good spectra
```

and the assertion fails. I suggest moving l.177 of desi.py inside the if statement in l. 180 to avoid the crash.

ngbusca:
With this change, the code runs smoothly (although I haven't checked whether this change breaks the bookkeeping somewhere else).

sbailey:
fixed by PR #30

PR #30, opened 2017-07-12
patch for issue 29 (bricknames and targets)

ngbusca:
This PR patches issue #29. This solves this particular issue, but I don't know whether it creates other bookkeeping problems down the pipe.


ngbusca:
Here's a plot comparing z from redrock and z from DR12Q for the 142 (drq-confirmed) quasars in plate 4199. It seems to work ok.

image

PR #31, opened 2017-08-14
shared memory refactor

sbailey:
This PR refactors the multiprocessing parallelism to explicitly use shared memory rather than relying upon os.fork() to share global state correctly.  This is a step towards a pure-MPI version that would also use shared memory arrays wrapped by numpy arrays.  This has been tested with 50k targets over 100 deg2 from the 2% sprint, and for the first time the combination of MPI+multiprocessing worked well at scale.  We should still pursue a pure MPI version, but this provides a working reference state to relieve some of the urgency for the MPI refactor.

### Technical details for the record ###

When the spectra are originally read from disk, their flux, ivar, mask, and resolution data are put into `multiprocessing.RawArray()` shared memory arrays.  These are wrapped with `numpy.frombuffer()` to make normal numpy ndarray views of these buffers.  This enables the algorithm code to use them without any changes.

Now for the messy bit.  When those are passed to new multiprocessing processes, normally multiprocessing would see the numpy array and try to pickle it thus losing the shared memory benefit.  So I had to implement pack/unpack functions to remove the numpy wrappers, pass to the multiprocessing function as a plain shared memory buffer (RawArray), and then convert them back to the numpy-wrapped state for use.

Summarizing what I like:
  1. explicitly sharing memory rather than implicitly hoping that os.fork will get it right (when wrapped by Cray MPI, it doesn't...)
  2. algorithm code didn't have to change; parallelism is still contained in just 2 wrapper functions of the core algorithm code.
  3. it works (for now)

Summarizing what I didn't like:
  1. I had to change the fundamental data objects themselves (Spectra, Targets), though at least this was done in a way that didn't impact algorithm usage of them
  2. The pack/unpack for passing to multiprocessing is messy, but is at least isolated to the two `parallel_*` functions and not anywhere else.

I think the MPI version will use a modified version of Bad Thing (1) but may be able to avoid Bad Thing (2).  We should strive to maintain as much as Good Thing (2) as we can (not mucking up algorithm code with parallelism details).

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ Redrock change log
 0.4.2 (unreleased)
 ------------------
 
-* No changes yet
+* refactored multiprocessing parallelism to use explicit shared memory
 
 0.4.1 (2017-06-16)
 ------------------
--- a/py/redrock/dataobj.py
+++ b/py/redrock/dataobj.py
@@ -89,7 +89,10 @@ class Spectrum(object):
         self.wavehash = hash((len(wave), wave[0], wave[1], wave[-2], wave[-1]))
 
     def sharedmem_unpack(self):
-        '''TODO: document'''
+        '''Unpack shared memory buffers back into numpy array views of those
+        buffers; to be called after self.sharedmem_pack() to restore the object
+        back to a working state (as opposed to a state optimized for sending
+        to a new process).'''
         self.wave = sharedmem.toarray(self._shmem['wave'])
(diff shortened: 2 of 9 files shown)


sbailey:
Discussed structure with Julien and prepared for MPI version.  Merging this now as a fresh base for the MPI branch.  Will try to maintain both multiprocessing and MPI versions for now.

rainwoodman:
this sounded very similar to sharemem.

I was using mmap to directly create the segments rather than using python's
raw array. raw array had some interesting memory management that creates
leaks from time to time.

PR #33, opened 2017-09-05
Further MPI improvements

tskisner:
This does some cleanup of the MPI code and unifies the rrdesi() function so that it works for both MPI and multiprocessing.  This also uses the MPIShared class unmodified from the upstream source.  As an added benefit, this means that redrock using MPI can scale to more than one node for a single spectral group / brick.

I did not want to merge this to the existing mpi branch without someone taking a look.  @julienguy, let me know if this looks ok.  I tested it on a couple small spectra files to confirm that the multiprocessing and MPI cases agree and that they also give the same results as the current mpi branch.

### DIFF
--- a/bin/rrdesi
+++ b/bin/rrdesi
@@ -4,4 +4,4 @@
 Redrock for DESI
 """
 from redrock.external import desi
-desi.rrdesi_multiprocessing()
+desi.rrdesi()
--- a/bin/rrdesi_mpi
+++ b/bin/rrdesi_mpi
@@ -3,5 +3,23 @@
 """
 Redrock for DESI
 """
+import os
+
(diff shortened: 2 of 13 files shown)


julienguy:
Merging this despite travis test fail (because merging to branch, not master).
 - Tested successfully on cori interactive session
 - I see you commented out some messages. It would be nice to bring some back in to be able to see the progress of the redshift fitting.
 - It would be useful to be able to run the code without mpi4py installed. A test on a machine intentionally without mpi4py installed crashes because of the import in sharedmem_mpi.py . There must be a way to avoid this, a lot of users would like it.

tskisner:
Thanks!  deleting branch now.

PR #34, opened 2017-09-14
Implement alternate MPI domain decomposition

tskisner:
This branch contains lots of work by Julien, Stephen, and some cleanup by me.  I have tested that the results agree between a handful of spectra files using both multiprocessing on one node and MPI on several nodes.  Opening this pull request for further comments / modifications / cleanup.  In particular, the travis failures complain about not finding mpi4py, which is not required.  So there may be a fix needed for that.

### DIFF
--- a/bin/rrdesi_mpi
+++ b/bin/rrdesi_mpi
@@ -19,7 +19,7 @@ else:
 from redrock.external import desi
 
 if have_mpi:
-	desi.rrdesi(comm=MPI.COMM_WORLD)
+    desi.rrdesi(comm=MPI.COMM_WORLD)
 else:
-	print("MPI not available, running serially")
-	desi.rrdesi()
+    print("MPI not available, running serially")
+    desi.rrdesi()
--- a/bin/rrdesi_mpi
+++ b/bin/rrdesi_mpi
@@ -1,0 +1,7 @@
+#!/usr/bin/env python
+
+"""
+Redrock for DESI
+"""
+from redrock.external import desi
+desi.rrdesi_mpi()
--- a/bin/rrdesi_mpi
+++ b/bin/rrdesi_mpi
@@ -3,5 +3,23 @@
 """
 Redrock for DESI
 """
+import os
+
+# MPI environment availability
+have_mpi = None
+if ("NERSC_HOST" in os.environ) and ("SLURM_JOB_NAME" not in os.environ):
+    have_mpi = False
+else:
+    have_mpi = True
+    try:
+        import mpi4py.MPI as MPI
+    except ImportError:
+        have_mpi = False
+
 from redrock.external import desi
-desi.rrdesi_mpi()
+
+if have_mpi:
+	desi.rrdesi(comm=MPI.COMM_WORLD)
+else:
+	print("MPI not available, running serially")
+	desi.rrdesi()
--- a/py/redrock/dataobj.py
+++ b/py/redrock/dataobj.py
@@ -144,8 +144,11 @@ class MPISharedTargets(object) :
         self._rank = self._comm.rank
         self._procs = self._comm.size
         
-        print('rank #%d : MPISharedTargets.__init__ starting'%self._rank)
+        #print('rank #%d : MPISharedTargets.__init__ starting'%self._rank)
 
+        if comm.rank==0 :
+            print('rank #%d : allocating a shared memory for targets '%self._rank)
+        
         self._n = 0
         self._dtype = np.dtype("float64")
         self._target_dictionnary = None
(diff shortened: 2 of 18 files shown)


tskisner:
When running fully packed (64 processes) on a KNL node with 96GB of RAM, if the input file is big enough this will exceed the per-process memory limit.  This is because we are reading the entire data set into the memory of one process rather than placing it directly into shared memory from the start.  I'll change this shortly.

tskisner:
We are punting for now on reading directly to MPI shared memory.  @sbailey , the unit tests now pass.

py/redrock/fitz.py
```diff
@@ -56,22 +60,24 @@ def minfit(x,y):
...
         for j in target_indices:
             targets[j].sharedmem_unpack()
             results = fitz(zchi2[j], redshifts, targets[j].spectra, template, nminima=nminima)
             #- return index with results so that they can be sorted
             qout.put( (j, results) )
     except Exception as err:
-        import traceback, sys
-        message = "".join(traceback.format_exception(*sys.exc_info()))
+        import traceback, syst
+        message = "".join(traceback.format_exception(*syst.exc_info()))
```
sbailey: is `syst` a typo, perhaps leftover from a search-and-replace?  Should be `sys`?

py/redrock/dataobj.py
```diff
@@ -118,8 +124,160 @@ def sharedmem_pack(self):
...
+                        scipy.sparse.dia_matrix((rdata, roffsets), 
+                        shape=rshape)) )            
+            targets.append(Target(id, spectra, coadd=coadd))
+
+        return targets
+
+
 class Target(object):
-    def __init__(self, targetid, spectra):
+    def __init__(self, targetid, spectra, coadd=None):
```
sbailey: Please document new `coadd` option in the docstring.  I haven't groked why this is necessary.

bin/rrdesi_mpi
```diff
@@ -0,0 +1,25 @@
...
+    except ImportError:
+        have_mpi = False
+
+from redrock.external import desi
+
+if have_mpi:
+	desi.rrdesi(comm=MPI.COMM_WORLD)
+else:
+	print("MPI not available, running serially")
+	desi.rrdesi()
```
sbailey: Gasp!  tabs have appeared!  I thought they were eradicated along with polio years ago.  Please search for other cases of tabs and fix those back to spaces.

py/redrock/external/desi.py
```diff
@@ -33,7 +36,8 @@ def write_zbest(outfile, zbest):
     zbest.meta['EXTNAME'] = 'ZBEST'
     zbest.write(outfile, overwrite=True)
 
-def read_spectra(spectrafiles, targetids=None):
+
+def read_spectra(spectrafiles, targetids=None, spectrum_class=None):
```
sbailey: Please document new `spectrum_class` option.  Consider changing the default to `SimpleSpectrum` instead of None so that read_spectra can be used from interactive analysis work with minimal hassle about needing to know which spectrum class to use.

py/redrock/external/desi.py
```diff
@@ -111,7 +115,8 @@ def read_spectra(spectrafiles, targetids=None):
 
     return targets, meta
 
-def read_bricks(brickfiles, trueflux=False, targetids=None):
+
+def read_bricks(brickfiles, trueflux=False, targetids=None, spectrum_class=None):
```
sbailey: Similar to `read_spectra`: document use of `spectrum_class` and consider making `SimpleSpectrum` the default.

py/redrock/dataobj.py
```diff
@@ -118,8 +124,160 @@ def sharedmem_pack(self):
...
+        self.nwave=wave.size
+        self.wave=wave
+        self.flux=flux
+        self.ivar=ivar
+        self.R=R
+        self.Rcsr = self.R.tocsr()
+        self.wavehash = hash((len(wave), wave[0], wave[1], wave[-2], wave[-1]))
+
+
+class MPISharedTargets(object):
```
sbailey: Please add comments about how this class is implemented.  The creation of `_target_dictionary` from an existing set of targets and then repackaging those with `_get_targets` into another set of targets seems a maintenance risk and/or candidate for future cleanup refactor, so let's get some comments in here while the ideas are still fresh.

tskisner:
Thanks for review, some of those classes were added by @julienguy , but I'll do my best to document them.  Shuffling spectra pieces around so that all the data for a given target is contiguous seems nice, but maybe it raises the question of why we don't do that in the first place (which I think is mainly driven by our storage as FITS HDUs).

julienguy:
I think I implemented all of Stephen's comments.

Issue #35, opened 2017-09-27
Large memory footprint when multiprocessing


ngbusca:
Top reports that all subprocesses have the same memory footprint as the main process. redrock is then filling up the memory of my local system (12 cores, 40 GB) when running on a boss plate (~1000 objects, ~4 exposures). I don't have this problem if I set ncpu to 1

Are subprocesses receiving the entire set of data instead of the fraction that the subprocess actually requires? If yes, can this be avoided?

ngbusca:
This gets solved by splitting targets instead or redshift range in the multiprocessing step. Will make this change in the rrboss branch.

julienguy:
Stephen has implemented a shared memory approach when running redrock with multiprocessing.
Maybe you could follow this for rrboss. An alternative is the MPI version that also runs on laptops. This would be useful to have if we plan to run a large BOSS/eBOSS data set at NERSC.

sbailey:
Both the MPI and multiprocessing versions of redrock are using shared memory to not copy all of the read-only input spectra to every process, or at least that is what is supposed to be happening.  I'm not sure how top reports that though; it might be double counting the shared memory.  Is the current state causing a problem (e.g. crashing with a memory error) or is it just that top looks like it is using a lot of memory and you are wondering if that could be optimized?  Could you give examples of your inputs and we could try it at NERSC too?

The calc_zchi2 scan parallelism over redshifts instead of targets was purposeful, since the most expensive step is evaluating the templates at each redshift, which can then be shared for all targets quickly.  Parallelizing that step over targets means that each one has to evaluate each template at every redshift and the performance will likely be much worse.

Details: Earlier versions of redrock relied upon multiprocessing to share memory without copying via POSIX fork "copy on write", but we found that to be problematic when wrapped with Cray MPI.  Sometimes it wouldn't copy the shared memory, and sometimes it would (and then die with out out of memory error).  Since PR #31 (shared memory refactor) I haven't had that problem.

The MPI version also uses shared memory, but we have a problem for large bricks since rank 0 reads everything in and then distributes to shared memory, and rank 0 is exceeding its share of {total_memory} / {number_of_ranks}.  Reading in chunks and distributing should work around that but requires a bit of an I/O refactor...

ngbusca:
But when targets are unpacked, aren't new arrays for flux, etc, being created thus replicating "locally" all the data objects?

ngbusca:
[...]

Issue #4, opened 2015-10-18
Write rrboss wrapper script


sbailey:
Write bin/rrboss wrapper script to read in idlspec2d spCFrame files and perform redshift fitting.  All BOSS-specific code should be isolated to this script and/or a redrock.survey.boss module.

ngbusca:
I've implemented a first version of this in the branch rrboss. It runs now only on individual exposures, with a slightly different signature as rrdesi:

`
rrboss --indir platedir --allspec --zbest zbest.fits --output rrdetail.h5  --spall spAll.fits
`
spAll is necessary to get the thing_id (mapped to targetid) from the fiber information. 

I'll run a few tests.

ngbusca:
@sbailey can I close this?

sbailey:
Fixed by #37 (thanks!).

PR #37, opened 2017-10-11
added capability for running (e)BOSS data (issue #4)

ngbusca:
I've written the module boss.py (similar to desi.py) to input (e)boss data. The calling signature differs from that of rrdesi:

`
rrboss --spplate <spPlate> --spall <spAll> [--use-frames]
`

It runs over one single spPlate at a time. spAll is required to find the target ids (thing_ids) from plate-fiberid. 

The resolution matrix is built from wdisp assuming a gaussian shape and ndiag = int(4*wdisp_max + 1) ("4-sigma").

If --use-frames is not set, the spectra are read from the spPlate (this corresponds to the "coadd" option in rrdesi).

If --use-frames is set, the code will find the individual exposures from the spPlate header and get the spectra from the spCFrame files, assumed to be in the same directory as the spPlate. Unlike for desi, the wavelength grid is different for each spectrum. Redrock will see this a having multiple bands so it doesn't pose a problem per se. However, when the target class tries to coadd the individual exposures it will simply recopy all the input data. **this needs to be addresses before merging**

The implementation of parallelism is refactored in this branch: each parallel job receives a fraction of the list of targets and scans over all redshifts while in master each job receives all targes and scans over a fraction of the redshifts. This change addresses issue #35. 

Objects with THING_ID == -1 (typically sky but also some ELG, ..., which were selected with different photometry) are given a TARGETID = PLATEMJDFIBERID. Do we want to reject sky fibers and/or calibration stars?


ngbusca:
I haven't tested with mpi (I don't know how to)

ngbusca:
I've added a flag to the Target class to avoid coadding. When the flag is set, the target class won't get the .coadd attribute.

sbailey:
This looks good except for the how THING_ID vs. (PLATE,MJD,FIBER) is mapped to TARGETID depending upon whether THING_ID is -1 or not.  Could we always use `targetid = plate*1000000000 + mjd*10000 + fiber` and ignore THING_ID (which I think would also mean we no longer need spAll either)?  It would be best to provide these conversions as two functions for end users to swap back and forth (double check these!)
```python
def platemjdfiber2targetid(plate, mjd, fiber):
    return plate*1000000000 + mjd*10000 + fiber

def targetid2platemjdfiber(targetid):
    fiber = targetid % 10000
    mjd = (targetid // 10000) % 100000
    plate = (targetid // (10000 * 100000))
    return (plate, mjd, fiber)
```

For rrdesi, I tested the parallelize over targets vs. redshifts at NERSC.  For small inputs (30 targets) the new form was 50% slower, but for large inputs (600 targets) they were within 3%.  That's fine.  At NERSC I don't see the memory difference that originally motivated this swap, but I'm willing to accept that the multiprocessing shared memory tricks work differently (worse) on other machines than at NERSC and on my laptop.

ngbusca:
@sbailey sure, I'll change the THING_ID as you propose. What tool do you use to check the memory footprint? Maybe I'm being fooled by top.

sbailey:
I based my memory estimate on the "maximum resident set size" reported by `/usr/bin/time rrdesi ...`.  Various tools have ambiguity in how they report shared memory usage by multiple processes, e.g.,
* (total memory) / (number of processes) so that the sum equals total memory used, but each process is reported as using less than it actually needs if it was running alone.
* (total memory used by the process) so that each process is reported as the amount that it needs if it was running alone, but has double counting across processes.

I haven't been able to find which one top is reporting.  And it is also possible that /usr/bin/time isn't counting the memory of spawned processes at all, in which case it wouldn't really be reporting what we want.

Bottom line: if you were having memory problems and this fixes it, ok, that's great.

ngbusca:
Refactored the TARGETID as proposed above. There are still two open questions:

* There's no special treatment for SKY fibers (i.e. they are passed to zscan as physical spectra). This might be useful as a null test. Should we instead remove SKY fibers?
* I added an option to the Target class to avoid the coaddition step. This option is used by rrboss to skip this step, since it results in duplication of the data set (for (e)BOSS data).

sbailey:
No special treatment for sky fibers is the same as what BOSS does, and is indeed useful as a null test.

Target option to avoid coaddition is fine.

Thanks for standardizing on (plate,mjd,fiber) <-> targetid mapping for rrboss.

Merging now.

PR #40, opened 2017-10-31
Read spectra files directly to MPI shared memory

tskisner:
This implements a new desi variant of the MPISharedTargets class which loads one or more spectra files directly into MPI shared memory.  The list of spectra for each target is then constructed with references into the shared memory.  For now, we have disable the creation of the CSR version of the resolution matrix.  This could be re-enabled and stored in shared memory if needed.

### DIFF
--- a/py/redrock/dataobj.py
+++ b/py/redrock/dataobj.py
@@ -425,10 +425,10 @@ class Target(object):
 
         if coadd is None:
             #- Make a basic coadd
-            compute_coadd(spectra, 
+            self.coadd = compute_coadd(spectra, 
                 spectrum_class=MultiprocessingSharedSpectrum)
         else :
-            self.coadd=coadd
+            self.coadd = coadd
 
 
     def sharedmem_pack(self):
--- a/py/redrock/dataobj.py
+++ b/py/redrock/dataobj.py
@@ -137,47 +137,86 @@ class SimpleSpectrum(object):
         self.wavehash = hash((len(wave), wave[0], wave[1], wave[-2], wave[-1]))
 
 
+def target_offsets(targets):
+    """
+    Construct a dictionary of buffer offsets.
+
+    This examines the sizes of the data for a list of existing targets and
+    creates a dictionary of offsets to use when packing these into a contiguous
+    memory buffer.  Offsets are computed for the wave, flux, ivar, rdata, 
+    roffsets and rshape data members.
+
+    The dictionary structure is
+    { target_id_1 : { "spectra" : [ {"wave" : [b,e] , "flux" : [b,e] ,  ....}
+    , ... ]} , "coadd" : [ ... ]},  target_id_2 : ... }, i.e. a dictionary of
+    targets identified by their id.
+    
+    Each target has two list of spectra labeled "spectra" and "coadd".  Each
+    spectrum in the two lists of spectra is a dictionary containing, for the
+    keys "wave", "flux", "ivar", "rdata", "roffsets" and "rshape", the begin
+    and end index of the array in the shared memory buffer.
+    
+    Args:
+        targets (list): the list of targets.
+
+    Returns:
+        - dictionary
+        - number of double precision elements in the memory buffer.
+    """
+    n = 0
+    dictionary = OrderedDict() # keep targets as ordered in list
+    for target in targets:
+        dictionary[target.id] = {}
+        for spectra_label, spectra in zip(["spectra", "coadd"], 
+            [target.spectra, target.coadd]):
+            spectra_list = list()
+            for spectrum in spectra :
+                spectrum_dict = {}
+                for label, asize in zip(["wave","flux","ivar","rdata",
+                    "roffsets","rshape"], [spectrum.wave.size, 
+                    spectrum.flux.size, spectrum.ivar.size, 
+                    spectrum.R.data.size, spectrum.R.offsets.size,
+                    len(spectrum.R.shape)]):
+                    spectrum_dict[label] = [n, n + asize]
+                    n += asize
+                spectra_list.append(spectrum_dict)
+            dictionary[target.id][spectra_label] = spectra_list
+    return dictionary, n
+
+
 class MPISharedTargets(object):
 
-    def __init__(self, targets, comm):
+    def __init__(self, size, offsets, comm):
         """
-        Place a list of targets into MPI shared memory.
-        This goes as follows.
-        1) The root process reads the targets, fills a dictionary containing
-        for each spectrum of each target, the indices of
-        the wave, flux, ivar, rdata, roffsets and rshape arrays in the shared
-        memory buffer ( see self._fill_target_dictionary for more details
-        on the format of this dictionary ).
-        2) The dictionary is broadcasted to all processes.
-        3) The root process allocates and fill the memory buffer with data.
-        4) All processes recreate the target list, using references to portions of 
-           the memory buffer for all wave, flux, ivar, rdata arrays.
+        Collection of targets in MPI shared memory.
+
+        This class wraps a buffer of MPI shared memory that is constructed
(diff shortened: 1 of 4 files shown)


julienguy:
- Successfully tested on cori haswell on ```/project/projectdirs/desi/datachallenge/dc17a-twopct/spectro/redux/dc17a2/spectra-64/327/32723/spectra-64-32723.fits``` (635MB)
   - running with ```srun -n 32 -c 2 rrdesi_mpi ``` on interactive node (with OMP_NUM_THREADS=1)
   - checked same redshifts as earlier version
   - memory usage smaller (21.1GB instead of 123GB at peak!), but this is still large (coadd issue?)
   - zscan is 2 to 3 times slower because of the sparse resolution matrix representation (DIA instead of CSR) 
 
- Successfully tested in prod (with desi_pipe...) on edison
  - there is also a factor 2 in CPU time
  - I get the same redshifts as master version (at least on two zbest files)

julienguy:
Because this is one of the most CPU intensive task, I'm afraid we cannot avoid the shared memory mapping of the CSR sparse matrices.

sbailey:
Agreed with @julienguy.  If the slowdown was 20-30% we could live it with, but 2-3x is too much.  Alas, time to sharedmemify `Rcsr` too.  `Rcsr.data`, `Rcsr.indices`, and `Rcsr.indptr` are the arrays to get into shared memory.

This branch also broke the `Target.coadd` attribute such that it no longer works for the multiprocessing case.  At a more general level, it is concerning to me that we are diverging into a state where that is even possible (multiprocessing is trying to use Target.coadd but MPI is not).  I hope that is just a debugging leftover from seeing the memory impact of coadds.

tskisner:
Ok, I believe the CSR matrix is correctly stored in shared memory.  @julienguy , let me know if this solves your performance regression.

tskisner:
I noticed that the "very large" 2.5GB spectra file does cause redrock to run out of memory now on an edison node.  Don't merge yet while I double check that the memory is not being copied on CSR construction.

tskisner:
Very strange, looking at the scipy source:
https://github.com/scipy/scipy/blob/master/scipy/sparse/compressed.py#L64
It sure looks like a compressed matrix can be constructed with the data, indices, and index pointers and copy=False and should just store a reference.  Yet the memory use explodes when I construct these from the shared memory buffer.  Will keep poking at it.

tskisner:
A small test shows that the **data** array is not copied when copy=False.  However, the index arrays would be copied if the dtype returned by scipy.sparse.sputils.get_index_dtype() was different than the input array dtype.  I had been using int64, but that function is likely returning int32 since our matrices are not very big.  Will make that change and see if it prevents copying of the the index arrays.

tskisner:
I disabled the CSR matrix construction and the high memory use persisted.  By examining the underlying pointers, I have found where memory is being copied in the construction of sparse dia matrix.  It looks like the constructor here:  https://github.com/scipy/scipy/blob/master/scipy/sparse/dia.py#L112 calls atleast_2d on the input data matrix (which is 2d, but reshaped from a 1d buffer), and determines that it must copy the data.  I'm going to change the native shape of the data stored in the shared memoy buffer to see if I can trick atleast_2d into avoiding a copy.  Sometimes python is just silly.

tskisner:
Just to document this problem in a standalone snippet, it looks like on construction the scipy.sparse.dia_matrix is copying the data array.  I replicated the process used in the dia_matrix constructor (see the fake_mat class), and in this case there is no copying (pointer values the same).  Compare this fake class with the constructor here:  https://github.com/scipy/scipy/blob/master/scipy/sparse/dia.py#L112

Somehow the scipy constructor results in a copy.  I will attempt to forcibly re-assign the scipy sparse data members.

```
import numpy as np
from scipy.sparse import dia_matrix

class fake_mat(object):
    def __init__(self, arg, shape=None, dtype=None, copy=False):
        data, offsets = arg
        self.data = np.atleast_2d(np.array(arg[0], dtype=dtype, copy=copy))
        self.offsets = np.atleast_1d(np.array(arg[1], dtype=np.int32, copy=copy))

# Create an initial dia_matrix

ddata = np.array([[1, 2, 3, 4]]).repeat(3, axis=0)
doff = np.array([0, -1, 2])

orig = dia_matrix((ddata, doff), shape=(4,4), dtype=np.float64, copy=False)
print(orig.toarray())

# Get the data vector and offsets from this pre-created matrix.  Make a copy
# just to be sure.

startdata = orig.data.copy()
startoff = orig.offsets.copy()

startdatap, rf = startdata.__array_interface__['data']
startoffp, rf = startoff.__array_interface__['data']
print("starting offsets: {} {} pointer = {}".format(startoff.shape, startoff.dtype.str, 
    startoffp))
print("starting data: {} {} pointer = {}".format(startdata.shape, startdata.dtype.str, 
    startdatap))

# Create a "fake" matrix that does the same things as the 
# dia_matrix constructor

fake = fake_mat((startdata, startoff), shape=(4,4), dtype=np.float64, 
    copy=False)

fakedata = fake.data
fakeoff = fake.offsets

datap, rf = fakedata.__array_interface__['data']
offp, rf = fakeoff.__array_interface__['data']
print("fake off: {} {} pointer = {}".format(fakeoff.shape, fakeoff.dtype.str, offp))
print("fake data: {} {} pointer = {}".format(fakedata.shape, fakedata.dtype.str, datap))

# Create a dia matrix

dia = dia_matrix((startdata, startoff), shape=(4,4), dtype=np.float64, copy=False)

diadata = dia.data
diaoff = dia.offsets

datap, rf = diadata.__array_interface__['data']
offp, rf = diaoff.__array_interface__['data']
print("dia off: {} {} pointer = {}".format(diaoff.shape, diaoff.dtype.str, offp))
print("dia data: {} {} pointer = {}".format(diadata.shape, diadata.dtype.str, datap))
```

tskisner:
I think I have finally isolated the problem which explains why my test above did not reproduce the scipy copy inside the dia_matrix constructor.  This line is the key:  https://github.com/scipy/scipy/blob/master/scipy/sparse/dia.py#L130
astype() seems to be copying the data if I specify "np.float64" to the constructor, even if the data is already in double precision.  The solution is to pass "None", and that skips the call.  I have this working in a modified copy of the test code above- now I'm working back up through the original code and making this change and testing.

tskisner:
This latest commit has the right numpy / scipy incantations to make sure that we can construct sparse matrices from raw buffers without copying.  This is one area where python is substantially more complicated than plain C/C++.  

I believe we also wanted to store the coadds in shared memory, correct?  That should only be important from a memory standpoint in the limit where we have very few spectra per target.  It probably means putting those coadds into separate buffers indexed by target, rather than the current buffers which are 2D buffers with one row per spectrum.

Also, even with the coadds disabled completely there is still substantial per-process memory allocation.  For example, when computing a subset of redshifts for all targets, this loop:
https://github.com/desihub/redrock/blob/tsk_mem/py/redrock/zscan.py#L164
essentially copies all of the flux and ivar data for all spectra in order to concatenate them.  Although the resolution matrix is larger, if every process on a node duplicates this flux and ivar data it does add up.  If we are really pushing memory limits, perhaps we can change the domain decomposition to parallelize over targets instead of redshifts, or prevent python data copies some other way.

sbailey:
This branch still runs out of memory on /global/cscratch1/sd/sjbailey/desi/dev/testspec/spectra-64-5299.fits, even when only using 8 ranks on cori.

The multiprocessing version also runs out of memory when using all 32 cores, but can process this file in 28.5 minutes when using 24 cores.  At minimum the MPI version should be able to do better than that before merging.

sbailey:
Closing this PR since it will be soon superseded by a PR from the refactor branch which solves the remaining memory problems in this branch by doing a deeper more MPI-centric refactor.  @tskisner please delete the tsk_mem branch from github when you are done with it.

Issue #38, opened 2017-10-18
add flexibility to the template io


ngbusca:
The current template io has the template classes hardcoded. I'd like to suggest that the template info (such as spectype, redshift range, etc) be moved to the input .fits files. This would be useful if a user wants to add a template class (e.g. QSO_BAL), she would simply add a .fits file with the appropriate information. The io would loop over the existing target.fits files and read the necessary target information from them.

sbailey:
Good idea.  Please implement that.  For backwards compatibility with the existing templates, I'd suggest that if a template does *not* have the keywords to specify the redshift range, spectype, etc., then the code should default to the parameters it uses now for the classes that it already knows about.  i.e. continue to work with the existing templates while also supporting the ability to add a new template file without having to change the code too.

londumas:
@sbailey, Seems to me that this issue is fixed. What do you think?

PR #41, opened 2017-10-31
Refactor io

ngbusca:
This PR addresses issue #38 

Summary of the changes:
 * the function read_template tries to find the redshift range from the template file. It defaults to the original hardcoded if it fails. This change will allow users to add templates for spectra different than the current hardcoded classes.
 * the function read_templates puts the templates in a dictionary instead of a list. This is useful when running redrock interactively: to find a template in a list (e.g. to make a plot) one needs a loop, while one cat directly get it in a dictionary (e.g. templates['QSO'])
 * changing the templates to a dictionary implies minor changes down the road: wherever it said "templates" now it says "templates.values()"

### DIFF
--- a/py/redrock/dataobj.py
+++ b/py/redrock/dataobj.py
@@ -62,6 +62,24 @@ class Template(object):
         flux = self.flux.T.dot(coeff).T / (1+z)
         return trapz_rebin(self.wave*(1+z), flux, wave)
 
+    def add_broadband(self,degree):
+        '''
+        Add polynomial terms to this templates.
+        The templates will be added as if they were PCA components.
+        '''
+        if hasattr(self,'broadband'):
+            raise Exception("Multiple calls to broadband not allowed")
+
+        self.broadband = True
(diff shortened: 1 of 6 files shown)


sbailey:
Thanks for this update.  The change of templates from list -> dict is a good usability improvement but will also break redrock.plotspec.PlotSpec (see https://github.com/desihub/tutorials/blob/master/redrock/RedrockPlotSpec.md for how to use that).  It's a simple fix and I keep thinking that I'll just do it but then run out of time at the end of the day, so if you get around to fixing that first, please do.

Please also run "python setup.py test" -- a bunch of unit tests were broken by the list -> dict change.  I think all of them are simple fixes, but please update those too.  Thanks.

ngbusca:
I've fixed the tests and PlotSpec. Note that PlotSpec doesn't work on boss targets, since they don't have a coadd attribute (PlotSpec uses the coadd to make plots), but we should probably fix that in a separate branch.

PR #42, opened 2017-11-15
set ivar=0 to masked pixels

julienguy:
PR to fix one issue that was considerably lowering the redshift efficiency for simulated data with cosmic rays: 
set ivar=0 to masked pixels. This might also change the BOSS/eBOSS redshift efficiencies.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,9 +5,11 @@ Redrock change log
 0.6.1 (unreleased)
 ------------------
 
-* No changes yet.
+* set ivar = 0 where mask != 0 (PR `#42`_)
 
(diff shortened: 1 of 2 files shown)


sbailey:
Good catch.  I thought we had fixed that awhile ago, but alas no.  Bonus points to @ngbusca who got it right for boss.  I also updated the desi code to set ivar=0 for masked pixels when reading brick files (deprecated, but we might as well get that right too as long as we still read bricks).

At the same time, we should keep this in mind.  It is possible that in the future we may mask certain pixels that might still be ok to include with redshift fitting, and we may need a more refined filter for what should get ivar=0 or not.

PR #44, opened 2017-11-18
fix issue 43

londumas:
fix issue when --templates is a folder

### DIFF
--- a/py/redrock/external/boss.py
+++ b/py/redrock/external/boss.py
@@ -245,7 +245,10 @@ def rrboss(options=None, comm=None):
         print('INFO: fitting {} targets'.format(len(targets)))
         sys.stdout.flush()
 
-        templates = io.read_templates(opts.templates)
+        if not opts.templates is None and os.path.isdir(opts.templates):
+            templates = io.read_templates(template_list=None, template_dir=opts.templates)
+        else:
+            templates = io.read_templates(template_list=opts.templates, template_dir=None)
 
         dt = time.time() - t0
         sys.stdout.flush()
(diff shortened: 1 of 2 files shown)


ngbusca:
Hi guys, sorry to jump in late, but this fix means that if the user sets a files instead of a directory for --templates the code will still run without any warning. The user might think the templates she provided were used for the fits, while it would actually be the default templates. 

The original code with a crash allowed the user to realize something was wrong with the arguments. 

I would prefer a fix where the code verifies that the argument of --templates is a directory, and outputs a warning and exists if it's not the case.

sbailey:
@ngbusca I might be misunderstanding this phrase, or misunderstanding the code, but I don't think this actually happens:

> The user might think the templates she provided were used for the fits, while it would actually be the default templates.

i.e. I don't see a case where the user would specify something that wouldn't get used.  I think it is useful to allow `--templates` to be either a single file or a directory as @londumas implemented, and trust the user to know whether they are pointing to a file or a directory and what will happen in each case.

The original purpose of the option was to allow the user to test against a single class, especially for debugging, without having to run all template classes or create special directories with just that one template file.  Perhaps that original option should have been called `--template` (not plural).  This PR expanded that to also allow it to be a folder to override `$RR_TEMPLATE_DIR` which is also handy.

To clarify, the current behavior is:
```
if --templates is provided:
    if --templates is a directory:
        read all template files in that directory
    else:
        use only that single file
elif $RR_TEMPLATE_DIR is set:
    read all templates in $RR_TEMPLATE_DIR
else:
    raise an error
```
I think this is reasonable (though you might still reasonably disagree; opinions are always welcome even after PRs have been merged).

ngbusca:
@sbailey yep, you are absolutely right, sorry for the noise!

Issue #48, opened 2017-11-27
CHI2 can be negative


londumas:
Among the 3,056,862 objects fit with rrboss 16,842 have CHI2<=0. Among them 18 have ZWARN=0.
They seem to be spectra with very few pixels.

Here is an example of a spectrum with only 13 valid pixels:
```
 target 3777552100761
znum spectype         z                zerr       zwarn       chi2     
---- -------- ----------------- ----------------- ----- ---------------
   0   GALAXY    0.684110380898 2.22265086481e-05     0 -0.679448604584
   1   GALAXY    0.963572171329 2.23922524482e-05     4   12.5200171471
   2   GALAXY    0.290527856756 1.25982246613e-05     0   18.2864579558
   3     STAR 0.000663565138503 1.45625809284e-05     4   84.2927164602
   4     STAR -0.00156789399554  8.7091695412e-06     4   91.2040898193
   5     STAR 0.000883054358289 1.30696754679e-05     4   97.3096341818
   8      QSO     3.49351411723 0.000142390448734     4   107.733959794
   9      QSO      1.5153153726  7.9556667662e-05     4   107.993852615
  10      QSO     1.23146576711 6.88101846302e-05     4   110.841642439
```

![capture d ecran de 2017-11-27 15-49-00](https://user-images.githubusercontent.com/3166436/33293571-7fe72662-d38a-11e7-917f-f6252172021e.png)

Here is the full list of these 18 objects:
```
plate, mjd, fiber, targetid
[[         3606         55182           547 3606551820547]
 [         3777         55210           761 3777552100761]
 [         3809         55533           646 3809555330646]
 [         4711         55737           817 4711557370817]
 [         4953         55749           136 4953557490136]
 [         5375         55973           400 5375559730400]
 [         6465         56279             6 6465562790006]
 [         7254         56625           511 7254566250511]
 [         7261         56603           714 7261566030714]
 [         7261         56603           970 7261566030970]
 [         7581         57195           426 7581571950426]
 [         7705         57332           782 7705573320782]
 [         8284         57419           761 8284574190761]
 [         8366         57432           507 8366574320507]
 [         8789         57358           500 8789573580500]
 [         4368         55559           507 4368555590507]
 [         4406         55858           501 4406558580501]
 [         4532         55559           761 4532555590761]]
```

londumas:
It seems redrock doesn't store the number of pixels used to fit. Is it true @sbailey?

sbailey:
[...]

PR #51, opened 2017-11-28
fix issue 48

londumas:
fix cases where chi2 is negative
https://github.com/desihub/redrock/issues/48

### DIFF
--- a/py/redrock/fitz.py
+++ b/py/redrock/fitz.py
@@ -51,6 +51,8 @@ def minfit(x,y):
     zwarn = 0
     if (x0 <= np.min(x)) or (np.max(x) <= x0):
         zwarn |= ZW.BAD_MINFIT
+    if (y0<=0.):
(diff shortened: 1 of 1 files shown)


julienguy:
I am not sure it is a good idea to just change the ZWARN.
ZWARN!=0 is interpreted as non trustworthy redshift, but I personally don't know the cause of those very small chi2 values. It doesn't necessary mean the redshift fit is wrong.

londumas:
@julienguy, for the minimum of the parabola to be less than zero, you need the chi2 scan to be very small. See example here (https://github.com/desihub/redrock/issues/48)

sbailey:
These negative chi2 arise when nearly all of the pixels are masked such that the chi2 is small and then the parabola fit dips below 0.  That case should already be flagged as LITTLE_COVERAGE, but ok to also flag this as a bad BAD_MINFIT.

Issue #49, opened 2017-11-28
track number of degrees of freedom used in chi2 calculations


sbailey:
redrock incorrectly doesn't track the number of pixels used in the chi2 vs. redshift calculations, which causes a bias when there are masked pixels such that the number of degrees of freedom varies with the redshift of the template.

Related: redrock also doesn't track that different templates have different numbers of eigenvectors.  This can make a difference for borderline cases of selecting between two almost-equally-good fits from different templates.

londumas:
If we track this number, we could use it to compare different fits with the chi2 probability. What do you think?

londumas:
@sbailey, do you want the number of pixels and of parameters to be stored in the zbest file ?

sbailey:
A correction to the original post: masked pixels *don't* cause a change to the number of degrees of freedom vs. redshift, because the data mask is always the same regardless of the redshift of the template being tested.  i.e. this isn't as big a problem as I originally thought.

None-the-less, it would still be handy to add 2 additional columns with the number of unmasked pixels used and the number of eigenvectors of the best fit template.

@julienguy and I just had a long conversation that filled a whiteboard with matrix algebra that might solve this in a more rigorous way, but in the meantime having that extra information could be useful.

londumas:
The templates have different number of eigenvectors. So setting the best fit as the one with the lowest chi2 is not perfect. Also comparing znum_i to znum_i+1 to a Delta chi2 of 9, is not correct.
template nb_vectors
-------------------------
Star-A, 5
Star-B, 5
Star-F, 5
Star-G, 5
Star-K, 5
Star-M, 5
Galaxy, 10
QSO, 4

londumas:
fixed.

PR #53, opened 2017-12-01
init of fix issue 49

londumas:
track number of degrees of freedom used in chi2 calculations
https://github.com/desihub/redrock/issues/49

Not finished.
@sbailey, could you tell me how to get the number of valid pixels if co-adding observations ?

### DIFF
--- a/py/redrock/dataobj.py
+++ b/py/redrock/dataobj.py
@@ -329,9 +329,6 @@ class Target(object):
         """
         self.id = targetid
         self.spectra = spectra
-        self.nvalidwave = 0
-        for s in self.spectra:
-            self.nvalidwave += (s.ivar>0.).sum()
 
         if not do_coadd:
             return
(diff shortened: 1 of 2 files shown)


londumas:
@sbailey, also how to compare a fit by QSO with nparam=4 and a fit of galaxy with nparam=10?

londumas:
@sbailey, I don't see where you replace the spectra by the coadded spectrum if the option is on.

sbailey:
It's a bit of a hack.  See [redrock.external.desi lines 280-283](https://github.com/desihub/redrock/blob/master/py/redrock/external/desi.py#L280-L283) where it (optionally) replaces the full `target.spectra` array with the `target.coadd` array as part of reading in the data.  The downstream code thinks it is working with the spectra, unaware that those got replaced with a coadd.

londumas:
@sbailey, ready to be merged.
Is ti ok for you?

Issue #47, opened 2017-11-27
SMALL_DELTA_CHI2 for different star templates


londumas:
A non negligible number of stars (from eye scan) are flagged as ZWARN!=0 because another template of star fits nearly as well.
Since the star classification is a continuum, it is normal that for some stars we get two compatible templates.
We could set a condition stating that if the second best template is also a star then SMALL_DELTA_CHI2=0

For example target=3586551810094: 
```
znum spectype         z                 zerr       zwarn      chi2    
---- -------- ------------------ ----------------- ----- -------------
   0     STAR (A) -0.000116860579997  1.9342047008e-05     4 4007.82786385
   1     STAR (B) -0.000114349129022 1.95406225734e-05     0 4015.84137551
   2     STAR (F) -0.000120274130256 1.95726752171e-05     0   4311.382544
   4      QSO      1.27647633093 0.000268442671122     0 8235.95752663
   5      QSO      1.15285765716 0.000289393887284     0 8548.70757259
   6      QSO     0.737408748348 0.000162942053437     0 8590.28038257
   7   GALAXY     0.947435678136 0.000110410356425     0 9288.04926471
   8   GALAXY    0.0341969933625 4.01854357817e-05     0  9587.7898065
   9   GALAXY     0.990873544233 0.000111714339476     0 9698.98343697
```

![capture d ecran de 2017-11-27 13-45-16](https://user-images.githubusercontent.com/3166436/33288359-37a4f840-d379-11e7-97f5-990eec897b5b.png)

londumas:
We could use the classification continuum of stars M, K, G, F, A, B, O.
For example if the first best template is A and the second is F or B, then SMALL_DELTA_CHI2=0, else =1

londumas:
If we want we would have to change the following lines:
```
            ii = np.where(tzfit['deltachi2'] < 9)[0]
            tzfit['zwarn'][ii] |= ZW.SMALL_DELTA_CHI2
```

sbailey:
Good catch.  It would probably be sufficient to just not flag SMALL_DELTA_CHI2 between templates that have the same spectype but different subtypes.  I.e. I don't think we need to do the special case bookkeeping of the order of the spectral types and which ones are "between" which others.

This same concept could also apply if we ever split out BGS / ELG / LRG / etc into separate templates with the same spectype but different subtypes.

londumas:
[...]

PR #54, opened 2017-12-02
fix issue 47

londumas:
Issue https://github.com/desihub/redrock/issues/47
SMALL_DELTA_CHI2 for different star templates

### DIFF
--- a/py/redrock/constants.py
+++ b/py/redrock/constants.py
@@ -1,2 +1,1 @@
-speed_light = 299792458. ## m/s
 max_velo_diff = 1000. ## km/s
--- a/py/redrock/constants.py
+++ b/py/redrock/constants.py
@@ -1,0 +1,2 @@
+speed_light = 299792458. ## m/s
+max_velo_diff = 1000. ## km/s
--- a/py/redrock/fitz.py
+++ b/py/redrock/fitz.py
@@ -230,7 +230,7 @@ def fitz(zchi2, redshifts, spectra, template, nminima=3):
         #- Skip this minimum if it is within constants.max_velo_diff km/s of a previous one
         #- dv is in km/s
         zprev = np.array([tmp['z'] for tmp in results])
-        dv = get_dv(z=zprev,zref=redshifts[imin])
+        dv = get_dv(z=redshifts[imin],zref=zprev)
         if np.any(np.abs(dv) < constants.max_velo_diff):
             continue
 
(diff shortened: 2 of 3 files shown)


sbailey:
Thanks for working on this, but I think we want a more generic solution that works for any of the redshifts, not just the 0th and 1st.  And following @djschlegel's comment in #47, we should consider doing it like BOSS and just not set SMALL_DELTA_CHI2 if two redshifts are within 1000 km/s even if they are different classes.  The exact number could be tuned and it isn't a specific physical reason, but it is pragmatically tuned for avoiding local false minima within the same template, and allowing two templates to agree on the redshift without clobbering each other with a SMALL_DELTA_CHI2.

londumas:
Do you mean something like this?
```
            for i in range(len(tzfit)-1):
                cond = (tzfit['zwarn'][i] & ZW.SMALL_DELTA_CHI2)!=0
                cond &= (tzfit['zwarn'][i+1] & ZW.SMALL_DELTA_CHI2)==0
                cond &= (tzfit['spectype'][i]==tzfit['spectype'][i+1])
                cond &= (tzfit['subtype'][i]!=tzfit['subtype'][i+1])
                if cond: tzfit['zwarn'][i] -= ZW.SMALL_DELTA_CHI2
```

djschlegel:
John- I think you want to be robust to the case where there could be many local minima within the arbitrary 1000 km/s range.  So, compare to the smallest chi2 outside that range.

londumas:
@djschlegel, where does this arbitrary `1000 km/s` come from?

djschlegel:
That's absolutely an arbitrary number.  It ought to be many sigma larger than the formal redshift errors  For BOSS, since our Poisson errors on the redshifts were ~80 km/s at the faint end, that was >10 sigma.  But if it's too large, then one would be ignoring false redshifts that we should be flagging as bad if another minimum is nearby -- in BOSS since we were searching galaxies across a redshift span of 1, that's a 0.3% effect.

One could argue whether this ought to be a decision that is made per-object.  For example, if a quasar had a redshift error of 500 km/s, should this number be larger for just that object?

londumas:
@sbailey, you define dv in py/redrock/fitz.py as:
```
dv = (scipy.constants.speed_of_light/1000.) * (redshifts[imin] - zprev) / (1.+redshifts[imin])
```
I define it in py/redrock/zfind.py as:
```
dv = scipy.constants.speed_of_light/1000.*(tzfit['z'][i]-tzfit['z'][i+1])/(1.+(tzfit['z'][i]+tzfit['z'][i+1])/2.)
```

what do you prefere? the min_z or the mean_z?

sbailey:
I'll actually propose a 3rd option:
```python
def get_dv(z, zref):
    '''returns velocity difference in km/s for two redshifts'''
    c = (scipy.constants.speed_of_light/1000.) #- km/s
    dv = c * (z - zref) / (1.0 + zref)
    return dv
```
Both places where we need `dv` should call that.  This form of the function has the advantage that `zref` could refer to a true redshift.

londumas:
@sbailey, good idea.
in both these cases would you set `zref`to z_min, z_max or z_mean ?

londumas:
@sbailey, where should we put this function?

sbailey:
Let's put it in `redrock.fitz.get_dv(z, zref)` and call it with zref = the redshift with the smaller chi2 and z = the redshift with the larger chi2.

PR #59, opened 2017-12-20
Add NUMTILE, NUMEXP to zbest output

### LINKED ISSUE #24: Add NUMOBS column to rrdesi zbest output
rrdesi zbest output should include a column NUMOBS including the number of observations that contributed to each result.

To decide:  The observing plan requests a certain number of observations, and then we might have to split a single observation over multiple exposures to achieve a normal level of S/N.  Do we want to store
1. Number of "observations"
2. Number of "exposures"
3. Both?

If (1) or (3), the concept of "observation" is somewhat ill defined -- two back to back exposures of the same tile on the same night probably count as one "observation".  But what about two exposures on back to back nights, which maybe were finishing off the requested S/N, or maybe were from a observing plan human override to get deeper exposures.  What about two exposures on the same tile separated by many nights?  I'll spin this into a desi-data email discussion and report back here.

sbailey:
Fixes #24 by adding NUMTILE and NUMEXP columns to zbest output.  Also propagates input fibermap from spectra file into output zbest file to allow matching targets/redshifts back to exposures.  Deprecates reading brick files.  Tested on 44k targets in 46 spectra files using minitest notebook.

This is the last (?) redrock change prior to the desimodules 17.12 release and prior to a major structural refactor by Ted to incorporate MPI more cleanly, so I plan to self merge after test pass (and after lunch...).

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -9,6 +9,7 @@ Redrock change log
 * API CHANGE: redrock.io.read_templates() returns dict not list (PR `#41`_)
 * set ivar = 0 where mask != 0 (PR `#42`_)
 * Add NUMEXP and NUMTILE to zbest output
+* Propagate input fibermap into output zbest
 
 .. _`#41`: https://github.com/desihub/desispec/pull/41
 .. _`#42`: https://github.com/desihub/desispec/pull/42
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -2,12 +2,13 @@
 Redrock change log
 ==================
 
-0.6.1 (unreleased)
+0.7.0 (unreleased)
 ------------------
 
 * Allow templates to optionally include redshift range (PR `#41`_)
 * API CHANGE: redrock.io.read_templates() returns dict not list (PR `#41`_)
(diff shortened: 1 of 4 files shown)


londumas:
Have you tested that there are no conflicts with rrboss?

sbailey:
Good question.  I had not explicitly checked that, but there were some changes to the shared data structures so that was a good idea.  I now checked that `rrboss` still works on DR14 data.  I did *not* try to propagate NUMTILE or NUMEXP into rrboss output zbest files.  We could add NUMEXP for rrboss output too, but NUMTILE is specific to DESI (or we could include it with NUMTILE=1 for datamodel consistency).

For the record, I check it with
```
rrboss --spplate /project/projectdirs//cosmo/data/sdss/dr14/eboss/spectro/redux/v5_10_0/4444/spPlate-4444-55538.fits --zbest zbest-4444.fits -n 64
```

londumas:
@sbailey, I don't think we need to propagate NUMTILE or NUMEXP into rrboss. If the current branch works, perfect.

PR #65, opened 2018-01-10
desiInstall compatibility

### LINKED ISSUE #28: fix redrock integration with desiInstall
Fix etc/redrock.modules and possibly desiInstall so that "desiInstall redrock {version}" will work at NERSC.  Currently it installs a modules file that points to /project instead of /usr/common and doesn't install the templates.

Consider whether the templates should always be installed as a separate product, or installed into redrock itself.  Templates are kept as a separate github repo to keep big files away from the code, but redrock supports them either being installed into redrock/templates, or in some other location while setting the `$RR_TEMPLATES_DIR` environment variable.

For the purposes of today, I will install and cleanup redrock by hand at NERSC.

weaverba137:
This PR fixes #28 and generally brings redrock into line with the DESI infrastructure.  redrock will be integrated with Coveralls and ReadTheDocs.


weaverba137:
That's fine.  It won't be hard to make the dependency optional.  However note that in addition to the version string, desiutil also provides `test --coverage` and `get_version()` which allows setup.py to read the _version.py file without having to actually import the package itself (which is an awkward thing to do when you're trying to install the package).

sbailey:
Good.  The goal is that if desiutil is installed you can get those bonus features, but if it isn't available you can still do basic stuff like "python setup.py test" and "python setup.py install".

It is fine for things like desiInstall to use those features because it knows that desiutil is installed.  Also ok to include desiutil in the travis config and use those features in travis.

sbailey:
@tskisner has given green light for merging this and then dealing with any conflicts in #67 rather that doing it the other way around.  I'll merge this now, and then after the dust has settled with the #67 refactor I'll update it to make desiutil an optional dependency.

PR #69, opened 2018-01-18
Make desiutil optional

weaverba137:
This should allow the basic commands `python setup.py test`, `python setup.py build_sphinx` and `python setup.py install` to work even if desiutil is not installed.  I successfully ran `python setup.py test` and `python setup.py build_sphinx` in such an environment.


sbailey:
fixes #69

Issue #66, opened 2018-01-11
read SDSS stellar templates


londumas:
The idea is to be able to run on templates from SDSS. The file spEigenStar-55734.fits gives these templates.
A simple code allows to convert these templates into files readabl by redrock.

```
def convert():

    ###
    p = "spEigenStar-55734.fits"
    cat = pyfits.open(p)

    for s in range(123):

        subtype = cat[0].header["NAME"+str(s)].replace(":","")

        ###
        header = fits.Header()
        header['CRVAL1']   = cat[0].header["COEFF0"]
        header['CDELT1']   = cat[0].header["COEFF1"]
        header['LOGLAM']   = 1
        header['RRTYPE']   = 'STAR'
        header['RRSUBTYP'] = subtype
        header['RRVER']    = ""
        header['INSPEC']   = p
        header['EXTNAME']  = "BASIS_VECTORS"

        ###
        hdus = fits.HDUList()
        hdus.append(fits.PrimaryHDU(sp.asarray([cat[0].data[s,:]]), header=header))

        ###
        subtypeWrite = subtype.replace(" ", "").replace("/","")
        outfile = "Templates/rrtemplate-star-"+subtypeWrite+".fits"
        hdus.writeto(outfile, overwrite=True)
        print('RR: Wrote '+outfile)
```

londumas:
@sbailey, the only issue is what to do with ` ` (spaces) `/`, and `:` in their names.
I can leave ` ` in the subtype names, but not in the file name
I can leave `/` in the subtype name, but not in the file name
I can not leave `:` in the subtype name. The reason is that it collids with the basic `type:subtype` format

sbailey:
It looks like you just strip out any ` `(spaces), `/`, and `:`, which seems fine for comparison testing.

However, let's *not* write 123 different star templates into redrock-templates for standard usage.  It would be better to upgrade redrock to support a format like spEigenStar-55734.fits and know how to treat the N>>1 rows as individual templates.  

We might also use a hybrid approach: keep the PCA basis for normal BAFGKM stars, and then include a subset of spEigenStar-55734.fits for the stars that are not well modeled by the PCA.

londumas:
> It looks like you just strip out any (spaces), /, and :, which seems fine for comparison testing.

for testing it works, but I was asking beyond tests

[...]

Issue #74, opened 2018-01-19
`n_target` should be `n_targets`


londumas:
https://github.com/desihub/redrock/blob/d0a32603c9d0c26c792eca18a17629db83c09ff4/py/redrock/external/boss.py#L314

In the new master, rrboss is broken: `n_target` should be `n_targets`.
Why even creating `n_targets`? We could simply use `args.ntargets`

```
    n_targets = None
    if args.ntargets is not None:
        n_targets = args.ntargets

    first_target = None
    if args.mintarget is not None:
        first_target = args.mintarget
    elif n_targets is not None:
        first_target = 0
```

```
        if args.ntargets is not None:
            targets = targets[first_target:first_target+n_targets]
            meta = meta[first_target:first_target+n_targets]
```

sbailey:
Thanks for catching this (unfortunately I only tested rrdesi, not rrboss).  Please go ahead and fix (but make sure that all combinations work: only --mintarget, only --ntargets, or both --mintarget and --ntargets).

Related to #70 #71 and #73.

Issue #58, opened 2017-12-16
Add Ldwarf templates


sbailey:
Add a set of Ldwarf templates to redrock-templates.

See issue #57 with an odd spike in the galaxy redshift distribution.  Many of these are mis-identified Ldwarf galaxies based upon SDSS classification.

sbailey:
It would probably be easiest to add a new type of template for redrock containing archetypes with a single template per SPECTYPE/SUBTYPE, similar to BOSS stellar templates in idlspec2d/templates/spEigenStar-55734.fits .  That would enable adding N>>1 types without requiring a new template file for each one.  We could cherry pick the subset of BOSS archetypes that are not well-modeled by the existing redrock stellar templates.  This would require an update to `redrock.io.read_template` to recognize this new format and convert them into individual templates.  Algorithm code could also be streamlined to recognize that when there is a single template the coefficient can be solved with dot products instead of setting up a fill matrix just to solve for a single coefficient

londumas:
Fixed by PR https://github.com/desihub/redrock/pull/78. Tested using only rrtemplate-star-Ldwarf.fits of spEigenStar-55734.fits on a random sample of 10 plates.

sbailey:
Let's leave this open until Ldwarf templates have actually been added to redrock-templates.  This is waiting for a determination of how many other stellar archetypes need to be added.

sbailey:
Fixed in redrock-templates 0.4, where the increased M-star training sample diversity was able to adequately model these.  redrock PR #66 also provides the tools to convert SDSS Ldwarf templates to redrock format if needed later.

PR #78, opened 2018-01-22
Read sdss templates

londumas:
Allow to read SDSS templates from `http://www.sdss3.org/svn//repo/idlspec2d/trunk/templates/`.

`bin/rrconvertSDSStemplates -infile <SDSS_templates> --outpath <folder_where_to_save>` allow to transform the SDSS templates into a format readable by redrock. One file per template.

Change the formating `spectype:subtype` into `spectype:::subtype` for two reasons: some stellar templates have a `:` in their name, also `:` is hard to look for with grep since it is used for each loop or condition in Python.

Fixes issue https://github.com/desihub/redrock/issues/66
Fixes also issue https://github.com/desihub/redrock/issues/74
Fixes also issue https://github.com/desihub/redrock/issues/83
Fixes also issue https://github.com/desihub/redrock/issues/82
Fixes also issue https://github.com/desihub/redrock/issues/58

### DIFF
--- a/bin/rrconvertSDSStemplates
+++ b/bin/rrconvertSDSStemplates
@@ -31,7 +31,6 @@ def convert():
         name_mode = True
     else:
         nb_templates = 1
-        eigen_mode = True
 
     ###
     for s in range(nb_templates):
--- a/bin/rrconvertSDSStemplates
+++ b/bin/rrconvertSDSStemplates
@@ -69,12 +69,3 @@ def convert():
     return
 
 convert()
-
-
-
-
-
-
-
-
-
--- a/bin/rrconvertSDSStemplates
+++ b/bin/rrconvertSDSStemplates
@@ -2,13 +2,18 @@
 
 import sys
 import scipy as sp
+import optparse
 from astropy.io import fits
 
 def convert():
 
+    parser = optparse.OptionParser(usage = "%prog [options]")
+    parser.add_option("-i", "--infile", type="string",  help="input SDSS templates")
+    parser.add_option("-o", "--outpath", type="string",  help="output filename")
+    opts, args = parser.parse_args()
+
     ###
-    p = sys.argv[1]
-    cat = fits.open(p)
+    cat = fits.open(opts.infile)
 
     if len(cat)!=1:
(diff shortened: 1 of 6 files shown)


londumas:
fixes also issue https://github.com/desihub/redrock/issues/79

londumas:
Tested on plate spPlate-4444-55538.fits and works very well.
Some SDSS templates raise a `'edges must be within input x range'`.
This is linked to the lack of wavelength coverage in some SDSS templates.
spEigenQSO-55732.fits and spEigenStar-55734.fits work perfectly.

sbailey:
For the record: I'll merge this once tests finish, but let's *not* add any of these new templates to redrock-templates until we have confirmed how many of them are needed.  If it is more than 5, I'll request that we use a format similar to idlspec2d spEigenStar*.fits with multiple archetypes in a single file instead of N>>1 new files.

londumas:
@sbailey, I am not sure removing the `(...)` part of the name is a good idea. I would keep it
For example in spEigenStar-55734.fits, we have:
NAME10  = 'A4V (136729)'                                                        
NAME11  = 'A4V (97603)'

sbailey:
OK, good catch that the same stellar subtype can be repeated with multiple archetypes, thus needing the Indo-US ID to disambiguate.  However, I think this is further evidence that we really want a trimmed down spEigenStar-like template file, instead of trying to force one archetype per file and hitting issues like subtypes with characters that aren't convenient for Unix filenames, or including info in the subtype that isn't really a subtype.

I restored the Indo-US ID for now, but let's revisit how to proceed when we know how many of these we really need to include.  e.g. if we don't really need the two A4V archetypes, we could drop the Indo-US ID again...

londumas:
@sbailey, we don't need to replace `" "` or `"("` or ")" for `RRSUBTYP`.
We only need to replace it for:
`outfile = opts.outfolder+"/rrtemplate-"+spectype+"-"+subtype+".fits"`

sbailey:
OK, yet another try...  I've reverted to something similar to what you originally proposed, with the addition of INDOUSID and SDSSNAME keywords, and stripping parentheses from the filename but not the RRSUBTYPE.

Issue #39, opened 2017-10-30
add additive broadband parameters


ngbusca:
I'm opening this issue to discuss how to implement the additive broadband terms. The easiest seems to be to extend the PCA components in each template with polynomials of lambda_rest. The redrock will handle them completely transparently. I've done tests in branch refactor_io and it works very well. 

The caveat to this is that they would get convolved with the resolution, so physically it represents an additive component to the pre-convolved model instead of something that comes in after. In practice, a polynomial is sufficiently smooth so it might not make a difference.

julienguy:
Convolving any spectrum by the resolution is CPU costly. There must be a
way to directly pass the monomials to the fitter.

sbailey:
If we (re-)add optional broadband terms (additive or multiplicative), they should be in the observer frame per spectrum such that the same functional form is included at any redshift.  Including them as extra PCA components is easy, but it means that template comparisons at different redshifts sample different portions of the broadband polynomial functions, and it doesn't allow one camera to be screwed up while others are ok.

An earlier version of redrock did have this, but I took them out as part of a refactor since it does considerably clutter the code and we weren't using them at the time.  For comparison, an example snapshot of the state of affairs when it did have optional polynomials:

https://github.com/desihub/redrock/blob/a75d04d04049086f70bda020a1cf09e2fe55de85/py/redrock/zscan.py

In the current code, additive polynomials per spectrum would be included as extra columns of the `Tb` matrix created in [zscan.py lines 197-202](https://github.com/desihub/redrock/blob/master/py/redrock/zscan.py#L202).  WARNING: that is undercommented code setting up some messy linear algebra, so watch out...

ngbusca:
'If we (re-)add optional broadband terms (additive or multiplicative), they should be in the observer frame per spectrum such that the same functional form is included at any redshift. Including them as extra PCA components is easy, but it means that template comparisons at different redshifts sample different portions of the broadband polynomial functions, and it doesn't allow one camera to be screwed up while others are ok.'

I don't understand this paragraph. Redshfiting a polynomial is equivalent to redefining its coefficients, isn' it? So the functional form stays the same... maybe there's something obvious I'm missing?

ngbusca:
Oh, I had coadds in mind. I can see how it wouldn't work accross cameras where one would like to allow for discontinuities in the bb

londumas:
Fixed in PR https://github.com/desihub/redrock/pull/84.

sbailey:
[...]

PR #84, opened 2018-01-23
add broadbands

londumas:
Add an option to add broadbands to the templates. The parameter is the number of different broadband coefficients.
```
bb_deg_i = ( (w-<w>)/<w> )**i
```
Tested on eboss data and works well. It removes the sky fibers that were mostly previously flagged as galaxies.
Testing on desi

fix issue https://github.com/desihub/redrock/issues/39

### DIFF
--- a/py/redrock/external/boss.py
+++ b/py/redrock/external/boss.py
@@ -265,6 +265,9 @@ def rrboss(options=None, comm=None):
         "(the spCFrame files are expected to be in the same directory as "
         "the spPlate")
 
+    parser.add_argument("--broadband-degree", type=int, default=None,
+        required=False, help="Degree of broadband parameters (default is no broadbands)")
+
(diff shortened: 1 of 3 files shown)


londumas:
Works on desi data
We might want to set it to 0 or 1 by default instead of None.
We would need to study on some plates the best to do.

ngbusca:
@londumas it looks like you added the power laws as extra pca components (i.e. in the rest frame and before convolution by the reaolutiob)? I understood from issue #39 @sbailey and @jguy were against this kind of implementation...

ngbusca:
@londumas another issue is that when fitting individual exposures, while the coefficients of the pca are the same for all exposures of the same spectrum, but the coeffs of the power law might be different...

londumas:
@ngbusca, I understand better now the discussions in https://github.com/desihub/redrock/issues/39

PR #90, opened 2018-01-25
Add script to easily plot redrock

londumas:
Here is a script `bin/rrplot` that can be run inside a `ipython --pylab` to plot redrock results.
It allows to transform from a multiple line command (like in tutorial) to a one line command.
Works both with `desi` or `boss`
example:
```
ipython --pylab
%run <path_to_rrplot> --datatype boss --data <path_to_spPlate> -o <path_to_h5_file>
```

### DIFF
--- a/bin/rrplot
+++ b/bin/rrplot
@@ -4,52 +4,40 @@ import argparse
 import redrock
 from astropy.table import Table
 
-def plotboss():
+parser = argparse.ArgumentParser(description="Plot redrock results for"
+    " BOSS target spectra.")
 
-    parser = argparse.ArgumentParser(description="Plot redrock results for"
-        " BOSS target spectra.")
+parser.add_argument("--datatype", type=str, default='desi',
+    required=False, help="input data type: desi or boss")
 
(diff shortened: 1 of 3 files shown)


sbailey:
Good idea to write a wrapper script and easily support both boss and desi.  It appears that it only works when called from ipython but not when called from the command line, but I haven't figured out why.  I also tried turning this into a function that one could call from ipython without the %run, and that also didn't work.  It seems like plot window wasn't getting the keypress events.  But calling from ipython via %run does work as you indicated.

I made two cosmetic changes:
* option `--output` -> `--rrfile` (normally -o/--output is reserved for the output of this script, not for an input that was the output of a previous script)
* option `--data` -> `--specfile` (we've got so much "data", I wanted to be slightly more specific)

I wish we could figure out how to also get it to work from the command line, but we can go ahead and merge anyway if you agree with those new option names.

londumas:
@sbliley, there is a way to forbid the window from closing. This allows to run from the command line, but then the keys don't work.
Ipython is better, because the different keys are very usefull.
I agree with your changes

PR #93, opened 2018-01-27
add seed to rrstartemplate

londumas:
This PR adds setting the seed to the desisim.templates.STAR production.
This is important to be able to reproduce the templates.
Should we put the seed inside the fits file header?
Do you want me to add this to the other template generation (i.e. for galaxy)?
Also the current code produces only 1000 spectra.
This means that only around 150 spectra are used for the definition of each templates.
We could change the code to have 1000 for each class.

### DIFF
--- a/bin/rrgaltemplate
+++ b/bin/rrgaltemplate
@@ -91,6 +91,6 @@ hdus = fits.HDUList()
 hdus.append(fits.PrimaryHDU(model.eigvec, header=header))
 hdus.append(fits.ImageHDU(model.coeff, name='ARCHETYPE_COEFF'))
 
-hdus.writeto(opts.outfile, clobber=True)
+hdus.writeto(opts.outfile, overwrite=True)
 print('RR: Wrote '+opts.outfile)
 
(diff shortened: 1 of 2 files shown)


sbailey:
This is good, thanks.  Please go ahead and make the other changes you suggested:
  * [ ] add seed to header
  * [ ] increase number of spectra generated so that each subtype has more templates going into the PCA
  * [ ] optional for this PR: add seed options for galaxies and QSOs

After those changes are merged let's make a new tag and re-generate the stellar templates to add to redrock-templates.

londumas:
@sbailey, I do that tomorrow. However, the QSO are simply read from eBOSS. Do you want me to change it so it read it from desisim?

londumas:
The first two points are done, half of the third is done.
Do you want me to transform `rrqsotemplate` from a conversion of SDSS templates to a generation of templates using `desisim.templates.QSO` ?

sbailey:
Looks good; let's merge as-is.  I forgot about "seed" not applying to QSOs since we just copy them from BOSS for now.  Rebuilding the QSO templates will be a separate project (and based on real data, not exclusively on PCA of desisim templates).

Issue #98, opened 2018-02-01
don't crash on NaN


sbailey:
From PR #97 discussion: redrock currently fails if the input spectra have NaNs.  They shouldn't have NaNs in the first place, but even if they do we should just filter them out as if ivar=0 for those wavelengths.
```
rrdesi /global/cscratch1/sd/jguy/sim2017/redux/month5/spectra-64/59/5925/spectra-64-5925.fits -o rr-64-5925.h5 --zbest zbest-64-5925.fits --mp 32
...
/global/common/software/desi/users/jguy/redrock/py/redrock/zscan.py:44: RuntimeWarning: invalid value encountered in multiply
  wflux = weights * flux
    Finished in: 5.1 seconds
  Finding best fits for template GALAXY
MP calc_zchi2: Traceback (most recent call last):
MP calc_zchi2:   File "/global/common/software/desi/users/jguy/redrock/py/redrock/zfind.py", line 46, in _mp_fitz
    t.template, nminima=nminima)
MP calc_zchi2:   File "/global/common/software/desi/users/jguy/redrock/py/redrock/fitz.py", line 219, in fitz
    results = Table(results)
MP calc_zchi2:   File "/global/common/software/desi/cori/desiconda/20180130-1.2.4-spec/conda/lib/python3.6/site-packages/astropy/table
/table.py", line 392, in __init__
    init_func(data, names, dtype, n_cols, copy)
MP calc_zchi2:   File "/global/common/software/desi/cori/desiconda/20180130-1.2.4-spec/conda/lib/python3.6/site-packages/astropy/table
/table.py", line 663, in _init_from_list
    self._init_from_cols(cols)
MP calc_zchi2:   File "/global/common/software/desi/cori/desiconda/20180130-1.2.4-spec/conda/lib/python3.6/site-packages/astropy/table
/table.py", line 728, in _init_from_cols
    .format(lengths))
MP calc_zchi2: ValueError: Inconsistent data column lengths: set()

/global/common/software/desi/users/jguy/redrock/py/redrock/zscan.py:44: RuntimeWarning: invalid value encountered in multiply
  wflux = weights * flux
/global/common/software/desi/users/jguy/redrock/py/redrock/fitz.py:60: RuntimeWarning: invalid value encountered in less_equal
  ii = np.where(np.r_[True, x[1:]<=x[:-1]] & np.r_[x[:-1]<=x[1:], True])[0]
```

londumas:
@sbailey, isn't that fixed?

[...]

PR #99, opened 2018-02-01
Check for Nan in flux and ivar of spectra files

julienguy:
set flux=0 ivar=0 to Nan or Inf values in spectra files when reading them.
checked it fixes the crash of one simulated spectra file in my sims.
now need to understand where this is coming from on the desispec side.

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -304,22 +304,16 @@ class DistTargetsDESI(DistTargets):
             for b in self._bands[sfile]:
                 extname = "{}_{}".format(b.upper(), "FLUX")
                 hdata = None
+                badflux = None
                 if comm_rank == 0:
                     hdata = hdus[extname].data[rows]
-                    # check for NaN here (should never happen of course)
-                    # we do a sum on waves to identify Inf values as well
-                    isnan=np.isnan(np.sum(hdata,axis=1))
-                    if np.sum(isnan)>0 :
-                        # NaN or Inf values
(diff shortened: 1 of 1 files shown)


julienguy:
This fixes issue #98

py/redrock/external/desi.py
```diff
@@ -322,7 +333,18 @@ def __init__(self, spectrafiles, coadd=True, targetids=None,
...
+                    # we do a sum on waves to identify Inf values as well
+                    isnan=np.isnan(np.sum(hdata,axis=1))
+                    if np.sum(isnan)>0 :
+                        # NaN or Inf values
+                        # We set flux to 0 for those
+                        # and will set ivar to 0 after
+                        # for entries with flux==0
+                        for i in np.where(isnan)[0] :
+                            hdata[i][np.isnan(hdata[i])]=0.
+                            hdata[i][np.isinf(hdata[i])]=0.
```
sbailey: I think this could be handled more simply with:
```python
bad = np.isnan(hdata) | np.isinf(hdata) | np.isneginf(hdata)
hdata[bad] = 0.0
```

I'm more concerned with the assumption downstream that `flux=0` data was actually bad data and should be set to `ivar=0`.  Real data will very very rarely be exactly 0, but that seems fragile.  Would be better to explicitly make a `badflux` mask like above, and then use that mask when loading the ivar HDU to set `ivar[badmask]=0.`

julienguy:
very well, I am redoing this.

Issue #85, opened 2018-01-23
Systematic error in quasar redshift due to Lya forest


londumas:
The following plot shows 1000 quasars generated with `desisim.templates.QSO()` and fitted with redrock.
Blue points give results with Lya forests, orange points give results without Lya forests
The idea is to test to convolve the first eigenspectra of quasar templates with the PDF (lya+lyb+...) and see if the systematic error can be removed or reduced.

![systematic_error_redshift_qso](https://user-images.githubusercontent.com/3166436/35306115-772a7b54-0059-11e8-9cf8-69b81e0dae03.png)

londumas:
Tested with a simple model, and seems to work well.
A more complete model should take the resolution of pixels into account, the presence of Ly-beta, Ly-gamma, ...

![systematic_error_redshift_qso_correction](https://user-images.githubusercontent.com/3166436/35308495-33bb7cfe-0065-11e8-8325-ab7a080d9722.png)


```
def mean_transmitted_flux_fraction(lRF,zpix):
     f = np.ones(lRF.size)
     w = lRF<1215.67
     f[w] = np.exp(-0.0023*(1.+zpix[w])**3.64)
     return f
```
Needs to modify redrock at two different places:

    - py/redrock/templates/DistTemplate
```
        for i, z in enumerate(myz):
            for k in list(self._dwave.keys()):
                lRF = self._dwave[k]/(1.+z)
                zpix = self._dwave[k]/1215.67-1.
                data[i][k][:,0] *= mean_transmitted_flux_fraction(lRF,zpix)
```
    - py/redrock/fitz/fitz()
```
            for k in list(dwave.keys()):
                lRF = dwave[k]/(1.+z)
                zpix = dwave[k]/1215.67-1.
                binned[k][:,0] *= mean_transmitted_flux_fraction(lRF,zpix)
```

londumas:
Here are two plots giving:
    - blue: master ( no correction )
    - orange (correcting for the transmitted flux fraction)
    - green (correcting for the transmitted flux fraction + capping the data weight blue-ward of the forest)
The black vertical lines give the redshift when the CIV and Lya lines appear in the data (i.e. lObs>3600.)

![systematic_error_redshift_qso_correction_capweights](https://user-images.githubusercontent.com/3166436/35360068-b266ffba-0119-11e8-8a9e-93a7e77d7285.png)

[...]

PR #104, opened 2018-03-09
Apply lyman transmission to pca

londumas:
Fix systematic error in quasar redshift for z>2.
It fixes it on the code side, but the templates side still needs some tests.

Here is a test using SIMQSO for generating spectra:
- `noCorrection` gives the current master
- `Correction+OldTemplates` gives the current branch
- `Correction+NewTemplates` gives the current branch with updated unabsorbed quasar templates

capture d ecran de 2018-03-09 15-58-47


londumas:
@moustakas, do you know how the parameters for the emission lines were extracted in SIMQSO? Unabsorbed or absorbed data? If it is absorbed data it could explain why updating the templates doesn't fix everything.

moustakas:
I don't know offhand, I'm sorry, and I can't dig into the simqso code base right now, unfortunately. I would pose the question [here](https://github.com/imcgreer/simqso/issues).

londumas:
Thanks @moustakas, the question is in https://github.com/imcgreer/simqso/issues/24.

moustakas:
For those of you interested in this thread (perhaps @andreufont @ngbusca @julienguy @profxj), @imcgreer posted a detailed answer to one of the questions posed by @londumas above:
  https://github.com/imcgreer/simqso/issues/24#issuecomment-372193756

londumas:
I have tested the effect on data of the branch versus the master without changing the templates.
It works very well. Here is the comparison of this branch versus all DR12Q redshifts for z>2.
We get a shift of -250 km/s and a tail at dv<-250 km/s against ZVI. This is very compatible with the effect found on the first plot of https://github.com/desihub/redrock/issues/85.

only_new_code_vs_dr12q

The issue with the new templates I produced and used to use is that the correction of the transmission produces a discontinuity in the Lyman-alpha emission line at lambda_RF=1215.67. This can be seen on the first eigen spectrum (in orange) on the following plot. This can be the cause of the shift in the green points of the first figure of this PR.

capture d ecran de 2018-03-13 17-35-35

sbailey:
@londumas can you post your redrock data files for DR12Q with the various methods?  e.g. somewhere at NERSC?

My continued confusion (correct me if any of these statements are wrong):
1. The new code was originally intended to work with templates that *don't* have LyA forest, by correcting them to have the mean-absorption for the redshift that is being tested
2. The new templates don't work well because the LyA correction during the template building creates a discontinuity at the LyA peak
    - it makes sense that those templates with the discontinuity don't work well
    - but why did the correction create such a big discontinuity?  It seems like the correction coefficients must be considerably different from what one would derive from the training sample used.  Why?
3. The old templates include LyA forest in some form because no correction was made while building them.
4. So why does new code (that tries to re-add LyA forest) work with old templates that already have LyA forest?
5. And why doesn't the new correction applied to the old templates create a different discontinuity that would be just as bad as the discontinuity created by applying the correction to the data while building a new template set?

I've also been puzzling over the "dv redrock vs. DR12Q" histogram, which would imply that Z_PCA, Z_MGII, Z_CIV, and Z_CIII are unbiased relative to each other, while Z_PIPE and Z_VI (visual inspection, not roman numeral 6) are the unusual ones.  Compare that to figure 7 of [Paris+ 2017](https://www.aanda.org/articles/aa/pdf/2017/01/aa27999-15.pdf) and accompanying text which would imply that Z_PCA, Z_MGII, Z_PIPE, and Z_VI are consistent and Z_CIV and Z_CIII are the unusual ones.

I haven't been able to reproduce figure 7 or the quoted numbers using DR12Q.fits itself, but I think the explanation lies in the z>2 cut vs. including all redshifts.  Plotting dv relative to Z_PCA vs. Z_PCA:
dv
* Z_PCA is most consistent with Z_MGII (red), which makes sense since the templates in Z_PCA were generated from a sample of QSOs with MGII redshifts
* There is a distinct z<2 vs. z>2 bias for Z_VI (black) and Z_PIPE (blue) relative to Z_PCA.  This is probably due to the BOSS spec1d pipeline having the same bias as we are chasing here for DESI.
* Z_PCA is mostly consistent with Z_CIV (green), with an additional bimodal bias of ~-1400 km/s above z>2.1 or so.  To a lesser degree Z_PIPE also shows that effect.

I thought the standard literature was that CIV was blueshifted relative to MGII, but this make it appear as a bimodal redshift failure driving that conclusion rather than a genuine bias.  I need to dig deeper into the older papers (e.g. Hewett & Wild 2010, or Shen+ 2016).

Bottom line: this PR seems to have redrock unbiased relative to DR12Q Z_PCA (good), but I don't understand why given the methodology and it isn't clear to me that we are getting the right answer for the right reason.

londumas:
@sbailey, thank you for the review.
Here are the zbest file for the different mode:
- Current redrock (master redrock and master redrock-templates)
`/global/cscratch1/sd/hdumasde/desi/rrboss/Add_WD_templates_from_desi__LDWARF_CARBON_from_SDSS/spplates/zbest.fits`
- This branch (PR redrock, master redrock-templates):
`/global/cscratch1/sd/hdumasde/desi/rrboss/Only_new_code_for_quasar_systematic/spplates/zbest.fits`
- This branch with "unabsorbed" quasar template (PR redrock, master redrock-templates, new "unabsorbed" qso template):
`/global/cscratch1/sd/hdumasde/desi/rrboss/New_qso_templates_from_data__corLYA/spplates/zbest.fits`

sbailey:
I updated my last plot to correct the labels.  The vertical axes are e.g. 3e5 * (Z_CIV - Z_PCA) / (1 + Z_PCA), such that for z>2:
* Z_PIPE and Z_VI are systematically greater than Z_PCA
* Z_CIV has a bi-modal blob at lower redshift than Z_PCA

I checked whether those are BALs -- many are, but certainly not all:
dv_civ-pca_bal
(the bottom histogram is normalized to area 1 for both BAL and non-BAL so that they are of similar size; focus on the distribution not the overall counts)

@julienguy and @londumas had a good idea to check the consistency of the stacked spectra for MgII and CIV for z<2 vs. z>2 for various flavors of redshift.  Please post here when that is ready.  In the meantime we're cleaning up other software release items so I'm leaving this open until the last minute in case new insights come in...

londumas:
Here are the result of stacking the spectra according to the different redshift available.
It is hard to get a definitive conclusion.
capture d ecran de 2018-03-21 11-00-34
capture d ecran de 2018-03-21 10-59-33
capture d ecran de 2018-03-21 10-58-01
capture d ecran de 2018-03-21 10-57-18
capture d ecran de 2018-03-21 10-56-18
capture d ecran de 2018-03-21 10-55-52

julienguy:
It is not totally obvious from looking at the figures. A normalization of the line amplitudes would help. However, some comments ...

1) redrock master does not show any significant shift of the CIV line between z<2 and z>2.
This is not surprising because redrock master does not correct for Lya transmission, and the redrock templates (coming from BOSS) have the average Lya line affected by Lya transmission at z ~2.5.
So, it is understandable that there is no bias **on average** over the redshift range where the Lya line is visible in the QSO spectra.

 2) rerock new code old templates seems to have a shift of the CIV line between z<2 and z>2.
This is not surprising because the Lya transmission is applied twice (once in the template creating that are based on uncorrected data, and once in the redrock code)

 3) rerock new code new templates does not seem to show any significant shift of the CIV line between z<2 and z>2.
  This means that the code is doing what we are expecting.


The fact you have an offset at z>2 with QSO simulations performed with simQSO is for me an indication of a bias on the Lya line position in simQSO. This line position has been derived from BOSS z~2.5 data **without Lya transmission correction** if I understand correctly the discussion in imcgreer/simqso#24. The Lya line, partially absorbed at short wavelength, appears shifted to larger wavelength in the data, and so in simQSO. So an accurate redshift fitter, fitting quasar spectra simulated with simQSO, would systematically overestimate their redshift at z>2 when the Lya contributes.

londumas:
@julienguy, Thank you for your summary.
Do you think we should dig into SIMQSO to solve this issue and confirm that this PR and the new templates do the job?

sbailey:
Below is an updated plot made from the coadds from @londumas that might make it easier to compare the different methods.  Rows are the methods, columns are the consistency of the z<2 (blue) vs. z>2 (orange) for the 3 lines CIV, CIII, MgII.
redrock_zconsistency

Even though offsets still exist for new code + old templates (and may even be a little worse for CIII and CIV), let's go ahead and merge this as an algorithmic step in the right direction.  This needs to be followed up with:
* new redrock templates built with a LyA correction while not introducing a notch at LyA
* updates to simqso (or post-facto corrections) to not include mean LyA in those simulated spectra.

Issue #108, opened 2018-04-03
redrock is stuck if LinAlgError: Singular matrix


londumas:
In `py/redrock/_zscan.py` at line `zcoeff[:] = np.linalg.solve(M, y)` if `M` is not positive definite the code gives this error, continues with the different fit but then gets stuck at 90%:

```
MP calc_zchi2: Traceback (most recent call last):
MP calc_zchi2:   File "<me>/Programs/desi/code/redrock/py/redrock/zscan.py", line 145, in _mp_calc_zchi2
    progress=qprog)
MP calc_zchi2:   File "<me>/Programs/desi/code/redrock/py/redrock/zscan.py", line 123, in calc_zchi2
    weights, flux, wflux, dtemplate.local.data[i])
MP calc_zchi2:   File "<me>/Programs/desi/code/redrock/py/redrock/zscan.py", line 76, in calc_zchi2_one
    zchi2 = _zchi2_one(Tb, weights, flux, wflux, zcoeff)
MP calc_zchi2:   File "<me>/Programs/desi/code/redrock/py/redrock/_zscan.py", line 31, in _zchi2_one
    zcoeff[:] = np.linalg.solve(M, y)
MP calc_zchi2:   File "<me>/.local/lib/python2.7/site-packages/numpy/linalg/linalg.py", line 375, in solve
    r = gufunc(a, b, signature=signature, extobj=extobj)
MP calc_zchi2:   File "<me>/.local/lib/python2.7/site-packages/numpy/linalg/linalg.py", line 90, in _raise_linalgerror_singular
    raise LinAlgError("Singular matrix")
MP calc_zchi2: LinAlgError: Singular matrix
```

To test, send:
```
rrboss --spplate /uufs/chpc.utah.edu/common/home/sdss07/ebosswork/eboss/spectro/redux/v5_10_9/10432/spPlate-10432-58136.fits
 --zbest zbest-10432-58136.fits --output rrdetail-10432-58136.h5
 --ntargets 1 --mintarget 479
```
i.e. this issue is with plate, mjd, fiber = 10432, 58136, 483

londumas:
One of the object giving such an error (plate, mjd, fiber = 10432, 58136, 483) has only 12 valid pixels.

londumas:
A simple solution is to do a try/except:
```
    try:
        zcoeff[:] = np.linalg.solve(M, y)
    except:
        return -9999.
```
then set one of the ZWARNINGFLAG. What do you think @sbailey?

PR #109, opened 2018-04-04
fix when matrix is not positive definite

londumas:
fix when matrix is not positive definite by catching the error and setting the chi2 to nan.
The rest of the code set then the BAD_MINFIT ZWARNING flag
fix issue https://github.com/desihub/redrock/issues/108

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,14 +5,16 @@ redrock Change Log
 0.10.2 (unreleased)
 -------------------
 
-* No changes yet.
+* Catch LinAlgErrors from bad input data (PR `#109`_).
+
+.. _`#109`: https://github.com/desihub/redrock/pull/109
 
 0.10.1 (2018-03-30)
 -------------------
 
 * Default QSO redshift range 0.05-4.0 instead of 0.5-4.0 (PR `#107`_).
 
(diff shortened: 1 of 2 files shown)


londumas:
The code raise the following error, I am investigating.
```
<me>/Programs/desi/code/redrock/py/redrock/fitz.py:61: RuntimeWarning: invalid value encountered in less_equal
  ii = np.where(np.r_[True, x[1:]<=x[:-1]] & np.r_[x[:-1]<=x[1:], True])[0]
```

ngbusca:
@londumas in case it helps... this means that you are comparing against nan, those comparisons will give `False`

sbailey:
NaNs are often a pain; it might be better to just return a magic number like 9e99 instead.

londumas:
@sbailey, I don't see where the error comes from.

sbailey:
@londumas I made a small change to catch only `np.linalg.LinAlgError` instead of all exceptions.

@ngbusca you are right that this case comes from a target with two few pixels.  It does get flagged with the zwarn LITTLE_COVERAGE flag downstream, but in the current code it doesn't realize that it has too few pixels until the end instead of the beginning.  Better would be to to a pre-flight check to not bother trying to fit targets with too few pixels, but in the meantime this PR adds some robustness.

PR #114, opened 2018-04-27
Improve spectra reading speed

sbailey:
This PR significantly improves the speed of reading spectra files by `DistTargetsDESI`.  The largest improvements were from:
* faster coadd by stacking resolution matrix diagonals and then creating a single `scipy.sparse.dia_matrix` instead of creating and stacking multiple `dia_matrix` objects.
* Only pre-create the CSR sparse format cache upon request (needed for fast running of redrock, but not for loading spectra with redrock, e.g. for visualization).

Testing on a file with 4134 exposures of 1694 targets, the read time went from 1m 27s to 15.5s.  I also verified that the Rcsr caching change didn't change the overall runtime of rrdesi for both MPI and multiprocessing versions.

### DIFF
--- a/py/redrock/external/boss.py
+++ b/py/redrock/external/boss.py
@@ -64,9 +64,9 @@ def write_zbest(outfile, zbest):
     hx.writeto(outfile, overwrite=True)
     return
 
-
+### @profile
 def read_spectra(spplate_name, targetids=None, use_frames=False,
-    fiberid=None, coadd=False):
+    fiberid=None, coadd=False, cache_Rcsr=False):
     """Read targets from a list of spectra files
 
     Args:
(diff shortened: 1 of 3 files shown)


londumas:
@sbailey, could you have a look at `boss` reading method? Could you apply the same improvements?

sbailey:
@londumas good idea.  I updated how the resolution matrix was constructed in `redrock.external.boss.read_spectra` and added a `cache_Rcsr` option, speeding up reading an spPlate file from 8.8 sec to 2.0 sec.

londumas:
Thank you @sbailey

londumas:
got the following error:
```
Proc 0: Traceback (most recent call last):
Proc 0:   File "<path>/redrock/py/redrock/external/boss.py", line 377, in rrboss
    cache_Rcsr=True)
Proc 0:   File "<path>/redrock/py/redrock/external/boss.py", line 214, in read_spectra
    targets.append(Target(targetid, spectra, coadd=coadd, meta=tmeta))
Proc 0:   File "<path>/redrock/py/redrock/targets.py", line 127, in __init__
    self.compute_coadd()
Proc 0: TypeError: compute_coadd() takes exactly 2 arguments (1 given)
```

londumas:
fixed, Thank you @sbailey.

Issue #115, opened 2018-05-01
ZWARN=1024 in high-z QSOs


sbailey:
In runs with latest master of desisim / desispec / specter, ~half of the low redshift QSOs have ZWARN=1024 (BAD_MINFIT).  See cell 75 of
https://github.com/desihub/desitest/blob/master/mini/minitest.ipynb

I'm pretty certain that this was not the case for the 18.3 reference run, but I'll have to dig out the minitest commit that contains that (and then tag for 18.3...).  So I think it is a recent problem, but either way it is something that needs to be fixed.

londumas:
Could you give an example of such a QSO?
Is that the reason why there is a gap in the QSO plot of cell 73, or it is another issue?

sbailey:
Correction: it is the high redshift QSOs that have a problem, not the low redshift ones (in the QA table, QSO_L means Lyman-alpha QSO, not Low-z QSO...).

The redshift gap is due to a mis-match in coverage in the mocks and is unrelated.

Files are /global/cscratch1/sd/sjbailey/desi/dev/end2end/spectro/redux/mini.  Example (healpix,targetid):

```python
from astropy.table import Table
import healpy
zcat = Table.read('zcatalog-mini.fits')
isqso = zcat['SPECTYPE'] == 'QSO   '  #- with spaces, sigh
ishiz = zcat['Z'] > 2.1
is1024 = zcat['ZWARN'] == 1024
print(np.count_nonzero(isqso & ishiz & is1024), np.count_nonzero(isqso & ishiz))

isbad = (isqso & ishiz & is1024)
ra = zcat['RA'][isbad]
dec = zcat['DEC'][isbad]
targetid = zcat['TARGETID'][isbad]
healpix = healpy.ang2pix(64, ra, dec, nest=True, lonlat=True)
list(zip(healpix, targetid))
```

```
[(5263, 288230398226337452),
 (5263, 288230398226337460),
 (5263, 288230398226337464),
 (5274, 288230398272474535),
 (5274, 288230398272474537),
 (5274, 288230398272474538),
 (5274, 288230398272474539),
 (5274, 288230398272474540),
...
 (5652, 288230399857921151),
 (5652, 288230399857921155),
 (5652, 288230399857921157),
 (5652, 288230399857921158),
 (5652, 288230399857921162),
 (5652, 288230399857921164),
 (5653, 288230399862115367),
 (5653, 288230399862115375),
 (5653, 288230399862115382),
[...]

PR #117, opened 2018-05-04
fix issue of Lya transmission in mpi

londumas:
Fix issue https://github.com/desihub/redrock/issues/115 where the Lyman-alpha transmission was not applied when running mpi.
However I don't know enough mpi to know if we could fix this in a more computing time efficient way.
Tested and works on:
```
srun -A desi -t 30:00 -C haswell --qos interactive -N 1 -n 32 -c 2
rrdesi_mpi /global/cscratch1/sd/sjbailey/desi/dev/end2end/spectro/redux/mini/spectra-64/52/5263/spectra-64-5263.fits
--output redrock-64-5263__mpi_test.h5
--zbest zbest-64-5263__mpi_test.h5
--targetids 288230398226337452,288230398226337460,288230398226337464
```

### DIFF
--- a/py/redrock/templates.py
+++ b/py/redrock/templates.py
@@ -300,12 +300,12 @@ class DistTemplate(object):
             for z in myz:
                 data.append(results[z])
 
-            # Correct spectra for Lyman-series
-            for i, z in enumerate(myz):
-                for k in list(self._dwave.keys()):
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks.  I confirmed that the zbest table and chi2 vs. redshift results are now identical for MPI and multiprocessing.

Issue #110, opened 2018-04-04
spectra without pixels are not in output


londumas:
As discussed with @sbailey, spectra without valid pixels or with too few pixels (<100) should still be in the output but with the `ZWARN=NODATA` or `ZWARN=LITTLE_COVERAGE`

londumas:
fixed.

PR #123, opened 2018-06-21
Add empty fibers and flag them

londumas:
fix issue https://github.com/desihub/redrock/issues/110 by saving empty fibers and flagging them.


londumas:
@sbailey, ready to be merged

Issue #121, opened 2018-06-13
QSO template can't be open with fitsio


londumas:
https://github.com/desihub/redrock/blob/714fee1b8fd083a50a3b622765fc9be2321a68cd/bin/rrqsotemplate#L38

This line is too long. Because of this it is put in two different lines using `CONTINUE`.
This is not supported by fitsio, only by astropy.io.fits.
```
INSPEC  = '/Users/sbailey/boss/products/Darwin/idlspec2d/trunk/templates/spEig&'
CONTINUE  'enQSO-55732.fits'
```

The returned error is:
```
  File "$HOME/Programs/desi/code/redrock/py/redrock/archetypes/bin/rrqsoarchetype", line 94, in <module>
    hdr = h['BASIS_VECTORS'].read_header()
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 1393, in read_header
    return FITSHDR(self.read_header_list(), convert=True)
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 3829, in __init__
    self.add_record(r, convert=convert)
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 3854, in add_record
    record = FITSRecord(record_in, convert=convert)
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 4161, in __init__
    self.set_record(record, convert=convert)
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 4195, in set_record
    self.set_record(record['card_string'])
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 4183, in set_record
    card=FITSCard(record)
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 4283, in __init__
    self.set_card(card_string)
  File "$HOME/.local/lib/python2.7/site-packages/fitsio/fitslib.py", line 4302, in set_card
    raise ValueError("CONTINUE not supported")
ValueError: CONTINUE not supported
```

sbailey:
fitsio is aware of this issue with some discussion about implementing it https://github.com/esheldon/fitsio/issues/118.

For redrock, we could trim the INSPEC keyword value to keep it short enough for the next rrtemplate-qso.fits update.  It should keep the *last* N characters (so to keep the filename) rather than the *first* N characters (which could end up chopping off the filename itself).

[...]

PR #124, opened 2018-06-21
save only the name of the templates

londumas:
save only the name of the templates.
Fix issue https://github.com/desihub/redrock/issues/121


weaverba137:
I would recommend trying Astropy 2 instead of Astropy 1.

sbailey:
I'm being a bit pedantic, but I swapped to using the more standard `os.path.basename` to get the filename instead of splitting on `/` and grabbing the last element.

londumas:
@sbailey, works for me.

Issue #25, opened 2017-06-13
tracking template provenance


sbailey:
Add provenance to rrdesi zbest*.fits and rr*.h5 output to track what version of the redrock-templates was used for the fit.  This might require adding versions/hashes to redrock-templates themselves.

londumas:
@sbailey, do you want me to fix this really quick?

sbailey:
Yes please.  Some options:
  1. Add versions to the template fits files themselves and store individual versions for each template in the redrock outputs.  Makes it a pain to know which git tag has that particular combination of versions.
  2. Parse the version number from `CHANGES.md` and be diligent about updating it with each merge.
  3. Add a `versions.txt` or similar file (and be diligent about keeping it up-to-date)
  4. like (2) or (3) but also add checksums for each template as a cross check.

londumas:
The proposed solution adds a `VERSION` key into the header of all templates and archetypes.
Then when reading them in redrock, the version is stored into `template._version`.
This is saved into zbest in the hdu=0 as a list of keyword, similar to spectra files:
```
RRVER   = '0.11.0.dev400'      / Redrock version                                
TEMNAM00= 'GALAXY  '                                                            
TEMVER00= '0.7...2018.06.21'                                                    
TEMNAM01= 'QSO     '                                                            
TEMVER01= '0.7...2018.06.21'                                                    
TEMNAM02= 'STAR:::A'                                                            
TEMVER02= '0.7...2018.06.21' 
...
```

sbailey:
Bringing some comments over from desihub/redrock-templates#7 to continue the conversation here in the ticket:

* Including the "0.7" part in each individual file implies that we have to update every file with every tag of redrock-templates, even if that file didn't change. This is particularly problematic for git versioning which doesn't diff binary files in an efficient way and will bloat the repo.
* 0.7...2018.06.21 is a non-standard version string format. Let's keep the repo tag names like "0.7" with dates in the CHANGES.md file like our other DESI repos.

I do like the idea of keeping the individual template file versions here, though this could also be done by adding FITS CHECKSUM and DATASUM keywords and using one of those.  It is also nice to have the overall "0.7" tag number to know what version of redrock-templates to checkout, but that should be tracked separately from the individual files so that they don't have to all be updated for every tag.

londumas:
[...]

PR #126, opened 2018-06-21
add version number when creating templates

londumas:
add version number when creating templates.
Maybe these different scripts to create template PCA dshould be in `redrock-templates/bin/`.
What do you think @sbailey?

Step in fixing https://github.com/desihub/redrock/issues/25


londumas:
@sbailey, ready to be merged

Issue #122, opened 2018-06-13
Maximum redshift of quasar


londumas:
The current redshift distribution:
https://desi.lbl.gov/svn/code/desimodel/trunk/data/targets/nz_qso.dat
is not zero at z>4., currently redrock stops at z=4. Should we fit up to z=5?
What do you think @nathalieGit and @sbailey?

![distribution_redshift](https://user-images.githubusercontent.com/3166436/41363281-bfbea150-6ef0-11e8-8c65-c7b5bd30a51f.png)

nathalieGit:
It’d be nice indeed to fit to z=5. The 1D analysis for instance does use quasars till as high as are available.


> On 13 Jun 2018, at 18:03, Hélion du Mas des Bourboux <notifications@github.com> wrote:
> 
> The current redshift distribution:
> https://desi.lbl.gov/svn/code/desimodel/trunk/data/targets/nz_qso.dat
> is not zero at z>4., currently redrock stops at z=4. Should we fit up to z=5?
> What do you think @nathalieGit and @sbailey?
> 
> 
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

-----------------------------------------------------------------------------
Nathalie Palanque-Delabrouille

Irfu/DPhP cosmology group
CEA, Université Paris-Saclay, 91191 Gif sur Yvette, France
tel: (+33) 1 69 08 39 62   
-------------------------------------------------------------------------------

andreufont:
Hi Helion,

On 13 June 2018 at 17:03, Hélion du Mas des Bourboux <
notifications@github.com> wrote:

> The current redshift distribution:
> https://desi.lbl.gov/svn/code/desimodel/trunk/data/targets/nz_qso.dat
> is not zero at z>4., currently redrock stops at z=4. Should we fit up to
> z=5?
> What do you think @nathalieGit <https://github.com/nathalieGit> and
> @sbailey <https://github.com/sbailey>?
>
It depends on what we want to use redrock for. If we want to use it for
something equivalent to the SDSS "quasar catalog", I'd say we want Redrock
to be able to fit redshifts all the way to z=7 or so... We don't want to
miss the most attractive quasars!

If we just want something for BAO + P1D analyses, then it is probably good
enough to stop at z=5.5 or so, although even then I would say that a quasar
at z=5.6 would be pretty cool for P1D.

Cheers,
Andreu

londumas:
[...]

PR #130, opened 2018-06-29
scan qso redshift up to z=6

londumas:
Fix issue https://github.com/desihub/redrock/issues/122 by scanning up to `z=6` for quasar.


sbailey:
Note: this will make QSO redshift z vs chi2 scans 20% longer, but ok.

Issue #138, opened 2018-08-21
zfind.py::zfind() can assign the chi2 curve of another target 


londumas:
When sending `rrboss` with the `targetids` keyword, with targets in decreasing order, we get a different result than if we send in increasing order. This means that `zfind.py::zfind()` mixes lists and dictionary.
You can see that `zz [15] ` are wrongly assigned, producing a difference of `deltachi2` for `TARGETID=7612569720936` between the two runs (see the two `print(zfit)` below).

An easy fix is to be sure that `targets` from `boss.py::read_spectra()` and from `desi.py::DistTargetsDESI()` is sorted by increasing TARGETID, but it seems important to understand what is happening at the `zfind.py::zfind()` level.
 

```
rrboss --spplate $BOSS_SPECTRO_REDUX/v5_10_10/7612/spPlate-7612-56972.fits
--output test.h5
--zbest test.fits
--targetids 7612569720936,7612569720925
```

```
print(zfit)
     chi2               coeff [10]                 z               zerr       zwarn             zz [15]              ... subtype ncoeff    targetid   znum   deltachi2   BRICKNAME
------------- ------------------------------ -------------- ----------------- ----- -------------------------------- ... ------- ------ ------------- ---- ------------- ---------
16661.9705066 2289.78961247 .. 567.744280331 0.104885485403 2.13185052243e-05     0 0.104755218635 .. 0.106282549174 ...             10 7612569720925    0 5447.11538499  7612-569
7488.72770112 559.999868368 .. 214.847865288 0.105109671317 2.92264867668e-05     0 0.103992344284 .. 0.105518620144 ...             10 7612569720936    0 1084.28524739  7612-569

```

 versus

```
rrboss --spplate $BOSS_SPECTRO_REDUX/v5_10_10/7612/spPlate-7612-56972.fits
--output test.h5
--zbest test.fits
--targetids 7612569720925,7612569720936
```

```
print(zfit)
     chi2               coeff [10]                 z               zerr       zwarn             zz [15]              ... subtype ncoeff    targetid   znum   deltachi2   BRICKNAME
------------- ------------------------------ -------------- ----------------- ----- -------------------------------- ... ------- ------ ------------- ---- ------------- ---------
16661.9630917 2289.78582169 .. 567.743768875 0.104885473345 2.13098093104e-05     0 0.103992344284 .. 0.105518620144 ...             10 7612569720925    0 5447.12279989  7612-569
[...]

PR #139, opened 2018-08-21
sort targetids

londumas:
Fix issue https://github.com/desihub/redrock/issues/138, by sorting targetids.
However, this doesn't fix the reason why `zfind.py::zfind()` can mix information in the first place.
Tested on rrboss and on desi-minitest.

### DIFF
--- a/py/redrock/external/boss.py
+++ b/py/redrock/external/boss.py
@@ -243,6 +243,8 @@ def read_spectra(spplates_name, targetids=None, use_frames=False,
 
     if targetids == None:
         targetids = sorted(list(dic_spectra.keys()))
+    else:
+        targetids = sorted(targetids)
 
     targets = []
     for targetid in targetids:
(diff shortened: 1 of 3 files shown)


sbailey:
Good catch on finding the bug.  Does this PR fix the underlying problem, or just make the outputs reproducible regardless of the input order of the targetids?  i.e. is it possible that zfind is still mixing target information and now just producing reproducibly wrong results for some inputs?  I'm not sure what is meant by "this doesn't fix the reason why zfind.py::zfind() can mix information in the first place."

londumas:
>  Does this PR fix the underlying problem

no
> is it possible that zfind is still mixing target information and now just producing reproducibly wrong results for some inputs?

yes, it is possible, but improbable.
Still we should check what is happening in `zfind.py::zfind()` that mixes things.

londumas:
Tested on `rrboss`, on `rrdesi` and on `rrdesi_mpi`. Works fine.

londumas:
@sbailey, ready to be merged.

PR #146, opened 2018-08-30
Skymask

sbailey:
This PR adds a pragmatically motivated mask of two sky lines based upon redshift failures that appear in the pixel level simulations but not the fast simulations:

5577: z=0.4958 puts [OII] on this line and [O I], [S III], and [N II] each line up with other sky residuals in the far red, while H-alpha has nearly redshifted off the detector and is nearly unconstrained.

9793.5 is a bright line on the red end of the Z camera where wavelengths solutions are extrapolated, resulting in frequent mis-subtractions.

The following histograms show the fitted redshift distribution with spikes from these lines.  redrock+archetypes helps a little bit, but the masking helps even more.  It isn't evident from this plote, but masking+archetypes is the best combination.

zhist

This could be followed up with a more detailed study of all sky lines, but this was a pretty easy improvement for pixel-level sims based upon just these two lines.

### DIFF
--- a/bin/wrap-redrock
+++ b/bin/wrap-redrock
@@ -413,6 +413,14 @@ def main():
     if args.plan:
         plan(args, comm=comm)
     else:
+        #- Create output directory if needed
+        if (args.outdir is not None) and ((comm is None) or (comm.rank == 0)):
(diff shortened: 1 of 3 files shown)


sbailey:
added --no-skymask option

PR #147, opened 2018-08-30
Run on one thingid across different plate

londumas:
Add the possibility to use all observations across all different plates and MJD to fit a single object

### DIFF
--- a/py/redrock/targets.py
+++ b/py/redrock/targets.py
@@ -40,7 +40,7 @@ class Spectrum(object):
         self.R = R
         self._Rcsr = Rcsr
         self._mpshared = False
-        self.wavehash = hash((len(wave), wave[0], wave[1], wave[-2], wave[-1]))
+        self.wavehash = hash((len(wave), wave[0], wave[1], wave[-2], wave[-1], R.data.shape[0]))
 
     @property
     def Rcsr(self):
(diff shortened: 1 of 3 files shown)


py/redrock/targets.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, wave, flux, ivar, R, Rcsr=None):
         self.R = R
         self._Rcsr = Rcsr
         self._mpshared = False
-        self.wavehash = hash((len(wave), wave[0], wave[1], wave[-2], wave[-1]))
+        self.wavehash = hash((len(wave), wave[0], wave[1], wave[-2], wave[-1], R.data.shape[0]))
```
sbailey: From this code, if different spectra of the same object have different numbers of diagonal elements in the resolution matrix then they will have different `wavehash` values and thus will not be coadded.  Is that the desired behavior?  That could be ok since they will still be fit as part of the same target for redshift fitting.  If you wanted, it would still be possible to coadd these using fancier indexing logic in `Target.compute_coadd`, but that may not be necessary.

londumas: It is the behavior that I want. It is ok if they are not coadded, as long as they are all used to fit the target redshift.

londumas:
Let me check if it works for desi.

londumas:
It works now on desi and eBOSS. Ready to be merged.
@sbailey, I have tried on eBOSS, in different modes and it works very well. I can fit 84 different observations of the same object, some are coadded, some are not, and it works very well!
This really proves the power of redrock!

Issue #60, opened 2018-01-08
add priors


ngbusca:
Is there a way to add a gaussian prior to the redshift fit or interest in implementing that? I volunteer to do it if the answer is "no, yes"

londumas:
@ngbusca, what is the "redshift fit" ?

ngbusca:
@londumas I mean adding a term to the chi2 of the form (z - z_ext)^2/sigma_ext^2 where z_ext and sigma_ext are externally provided to the code. 

I guess it could be added as a module that reprocesses the rrdetails, this way I wouldn't need to change anything within zscan. It doesn't need to be part of redrock, but it's practical to have it there.

londumas:
@ngbusca, sorry I was confused with another repository. I didn't see it was redrock.
I don't see when it could be useful to have a Gaussian prior. When would you see it useful?

ngbusca:
@londumas it could be useful if you have an external measurement of the redshift that is not very precise (like a visual scan or similar) and you want to improve it with redrock.

sbailey:
I think I'd prefer a solution that re-loads the chi2 vs. z from the hdf5 file, updates the chi2 with whatever prior is wanted, and then call fitz again to get the new set of minima.  Doing that may be messier than it needs to be, so it would be useful to write a brief tutorial-style demo of how that is done and then consider if there is something that would make it easier.

For consideration: do you really want a gaussian prior (that necessarily biases the answer toward the middle of that), or perhaps instead a flat redshift window with a penalty outside the window?

ngbusca:
@sbailey Thanks for the input. I'll write some proposal code and we can improve from there. I'll leave the option of the type of prior open to the user to choose (with explicit options for gaussian and boxcar and such that it's easy to add other functional forms).

moustakas:
Apologies for hijacking this thread -- as this should really be a separate issue -- but another valuable development for redrock would be to include the object photometry when constructing P(z).  This should help both with increasing delta-chi2 between redshifts of a a given spectral template type and with classification (i.e., P(z) across different spectral types).

sbailey:
implemented in #152 (thanks @londumas)

PR #152, opened 2018-09-20
add reading and applying redshift priors

londumas:
This PR adds a simple implementation of adding redshift priors to
redrock (https://github.com/desihub/redrock/issues/60).
This will allow QuasarNet to compute the precise redshift using redrock.
The input file can be created with the following code.
it needs a TARGETID, Z and SIGMA field,  extname = 'PRIORS'.
The code is using a Gaussian prior, but we could add an option to use other type of priors.

In this PR, there is also a small fix allowing archetypes to read PATH variables.

```
import scipy as sp
import fitsio

def main():

    ### load a plate
    path = '$DESI_WORK/rrboss/SDSS_redrock_archetypes_newDeltaChi2/spplates/7339/zbest-7339-56683.fits'
    h = fitsio.FITS(path)
    t = h[1]['TARGETID'][:]
    z = h[1]['Z'][:]
    s = 0.1*sp.ones(z.size)

    ### save
    out = fitsio.FITS('priors-7339-56683.fits','rw',clobber=True)
    data = [t,z,s]
    names = ['TARGETID','Z','SIGMA']
    extname = 'PRIORS'
    out.write(data,names=names,extname=extname)
    out.close()

    return

main()
```


londumas:
One think to decide is:
- Should we expect all TARGETID to be in the prior file
- Should we add 0. to the chi2 surface if the object is not in it?

londumas:
This is what the chi2 scan would look like.
A huge parabola with some minima. Would that be ok? Do we want a prior smoother, that goes to zero
on large scal. like `exp(gauss)` ?
chi2_scan_prior

londumas:
@ngbusca and @sbailey, this PR is ready to be reviewed:
It adds basically three things:
- add option to add Gaussian priors to the coarse chi2 scan
- add option to read the chi2 scan from a .h5 file, i.e. from an already ran redrock (go from 13m5 on a plate to 8m14)
- add reading and writing when path uses PATH variables

@sbailey, I will be happy to change some variable names if you don't like the name I gave them.

londumas:
Using the 11 plates with the largest number of objects from DR12Q, and using Z_VI from DR12Q as a prior with `dz = 0.01`, I have the following results when comparing to DR12_Superset.
The conclusion is that we remove cata-failure for quasars and improve by 2% the number of good targets.

No prior
```
                   ntarg   good  fail  miss  lost
GALAXY              1127  96.72  1.24  0.80  1.24
QSO                 4141  95.99  1.40  1.21  1.40
STAR                1035  93.24  1.74  3.00  2.03
```

with prior:
```
                   ntarg   good  fail  miss  lost
GALAXY              1127  96.98  0.98  0.53  1.51
QSO                 4141  97.87  0.05  1.57  0.51
STAR                1035  93.91  1.55  1.93  2.61
```

sbailey:
Looks good.  Please
* document in the docstring for the `Priors` class the format of the input file that it needs.
* add a test (or extend `test_zfind` in test_zscan.py) that calls `zfind` with a prior.  It doesn't have to be fancy; just a "does it run and not crash" test will help us keep it that way in the future.

@ngbusca does this give you what you would need to use QuasarNet + redrock together?

londumas:
@sbailey,
- I just added the doc, could you have a look and tell me if it is ok?
- I also added a unittest, it produces a fits file of prior and runs with it.

Let's wait for @ngbusca to give me an example of a file from QuasarNet, so we can see if we have to change some things. Maybe add a SPECTYPE keyword, to apply the prior only to a given SPECTYPE.

londumas:
@sbailey, Travis is missing fitsio to work.

sbailey:
Please update the test to use astropy instead of fitsio, since astropy is installed in the test environment already.  Thanks.

londumas:
@sbailey, ready to be reviewed and merged.

PR #155, opened 2018-09-27
Apply codacy recomendations 2

londumas:
Codacy is also an iterative process, it finds some new recommendations when some other are fixed.
This PR fixes the few remaining ones that pop up after the last PR was merged.
I also removed some trailing white-spaces.
Some remains, like not catching exceptions and some unused variable: `nx = len(x)` in `py/redrock/_rebin.py` and `nz = len(myz)` in `py/redrock/templates.py`. That might be worth looking at.


londumas:
@sbailey, all checked are finished. Ready to be merged.

PR #170, opened 2019-09-25
adds --no-mpi-abort option

### LINKED ISSUE #169: rrdesi comm.Abort() interferes with desispec pipeline TimeoutError
`redrock.external.desi.rrdesi` has a fairly blunt "fail early fail often" mechanism:
```python
...
try:
    [stuff]
except:
    [print traceback]
    if comm is not None:
        comm.Abort()
```

The original motivation for this was that when running rrdesi by itself, exceptions on one rank were causing that rank to hang while other ranks stayed alive waiting for it to catch up at `comm.barrier`, `comm.bcast`, etc. steps and rrdesi would never exit until killed (with the error possibly stuck in stdout/stderr buffering).  Printing a traceback and calling `comm.Abort()` at least got us an error message and exited as soon as the problem occurred.

However, `desispec.pipeline.run.run_task` wraps each task with a timeout signal handler so that if a task gets bogged down on something like an slow I/O issue, it doesn't waste the full jobtime before getting killed.  The problem?  rrdesi is catching that `TimeoutException` and calling `comm.Abort()` without passing it back up the chain for the pipeline to gracefully mark that task as failed and move on with the others.  This is also killing off all other tasks running at the same time.

A few options:
* special case `TimeoutError` and re-raise that and only that.
  * This would fix this particular case, but still leaves us vulnerable to other
    single-rank exceptions taking down an entire production instead of
    cleanly failing on just that single task.
* add a `no_mpi_abort` option to rrdesi that the pipeline could use to
  prevent it from calling `comm.Abort()`, and having it re-raise whatever
  exceptions it finds instead.
  * This would also work for the `TimeoutError` because it is raised on all ranks, but it leaves us susceptible to the original problem of a single-rank exception blocking the code while the other ranks wait forever at some MPI barrier.  When run by `desispec.pipeline.run.run_task` the other ranks would eventually be interrupted by a `TimeoutError` so maybe that's ok.
* other?

I think what we really want doesn't exist within MPI: a way for one rank to force an exception on another rank without the full brutality of an MPI abort taking down the entire global communicator.

sbailey:
This PR adds a `--no-mpi-abort` option to both `rrdesi` and `rrboss` to prevent them from aborting MPI if a rank raises an exception.  This fixes #169 .

The default behavior when running `rrdesi` by itself is to still abort MPI upon failure.  A separate PR in desispec will use `--no-mpi-abort` to avoid the abort and use the TimeoutError signal handler to end rrdesi instead so that it can cleanly kill a single task and move on with the others.

### DIFF
--- a/py/redrock/external/boss.py
+++ b/py/redrock/external/boss.py
@@ -349,6 +349,9 @@ def rrboss(options=None, comm=None):
     parser.add_argument("--use-andmask", default=False, action="store_true",
         required=False, help="uses and_mask values to set masked pixel's ivar to zero")
 
+    parser.add_argument("--no-mpi-abort", default=False, action="store_true",
+        required=False, help="Do not call MPI Abort upon failure of a single rank")
+
     parser.add_argument("--debug", default=False, action="store_true",
         required=False, help="debug with ipython (only if communicator has a "
         "single process)")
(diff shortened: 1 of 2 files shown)


sbailey:
tested at NERSC.  Merging.

PR #173, opened 2020-01-31
Allow string for input to read_spectra

### LINKED ISSUE #172: boss.read_spectra() should promote input filename to list
`redrock.external.boss.read_spectra()` takes a `list()` as input not a string.  Internally, it iterates over the items in the input filenames, but this operation is also valid on a string.   So for, example, it might try to open the file `'/'` when the input is `'/foo/bar.fits'`. This function should be able to detect that.

weaverba137:
This PR fixes #172.  I also removed a few Python 2 lines that shouldn't be needed any longer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,9 +54,9 @@ env:
         # to repeat them for all configurations.
         # - NUMPY_VERSION=1.10
         # - SCIPY_VERSION=0.16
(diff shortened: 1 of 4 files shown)


weaverba137:
Hmm, that's actually a documentation error, not necessarily a Python 3.5 dependency problem.  But should be easy to fix.

PR #186, opened 2021-02-05
use tempfile+rename for output

sbailey:
Rongpu found the hard way that redrock wasn't using temporary intermediate files for its output, leading to partially written files with the "final" name in the case of a job timeout.  This PR updates redrock to follow the pattern used in desispec: write a temp file and then rename it only after successfully writing the final file.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,8 @@ redrock Change Log
 0.14.5 (unreleased)
 -------------------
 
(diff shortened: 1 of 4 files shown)


sbailey:
oh yeah, Travis delays.  tests pass locally; I'm going to merge without waiting for Travis (which also reminds me to do the switch for travis -> github actions...)

PR #194, opened 2021-06-14
add top hat prior

echaussidon:


echaussidon:
Hi all, 

This PR is an update of the [#PR 152](https://github.com/desihub/redrock/pull/152). I add a new prior (tophat prior) and I update the prior class to choose which prior has to be used. (WARNING: the prior has to be the same for all targets in the same spectra file.

This information has to be set in the prior file : 

     import scipy as sp
     import fitsio

    def main():

    ### load a plate
    path = '$DESI_WORK/rrboss/SDSS_redrock_archetypes_newDeltaChi2/spplates/7339/zbest-7339-56683.fits'
    h = fitsio.FITS(path)
    targetid = h[1]['TARGETID'][:]
    z_prior = h[1]['Z'][:]
    sigma = 0.1*sp.ones(z.size)
    function = np.array(['tophat'] * z_prior.size) #need to be the same for every one

    ### save
    out = fitsio.FITS('priors-7339-56683.fits','rw',clobber=True)
    data, names, extname = [targetid, function, z_prior, sigma], ['TARGETID', 'FUNCTION', 'Z', 'SIGMA'], 'PRIORS'
    out.write(data,names=names,extname=extname)
    out.close()

    return

    main()

This PR has no impact on the main code, only used with prior file setting function = 'tophat'. The current behaviour (with a gaussian prior) is simply recovered setting function = 'gaussian'.

redrock_zxhi2

echaussidon:
@sbailey :  Here, my PR of RR for the re-run of RR that I proposed to build the QSO catalog for clustering

sbailey:
@echaussidon thanks.  Please add a file in doc/priors.rst defining the priors format.

> WARNING: the prior has to be the same for all targets in the same spectra file

Clarifying: You mean that the functional form of the prior has to be the same for all targets?  e.g. all tophat or all gaussian?  That sounds ok.  It looks like the code does support a different redshift and sigma (or tophat width) per target (if it didn't, that would be a real pain...)

echaussidon:
> Clarifying: You mean that the functional form of the prior has to be the same for all targets? e.g. all tophat or all gaussian? That sounds ok. It looks like the code does support a different redshift and sigma (or tophat width) per target (if it didn't, that would be a real pain...)

For each targets, the value of the redshift and the sigma can be fixed independently. The functional form will be the same for all targets. It was already like that. I take care to not modify the core of the code and adapt what I want to do with the form of the function in the prior.

echaussidon:
@sbailey : I add the doc in `doc/prior.rst`

Once the PR will be merged, I will check if my code work with the desi version of RR and develop the code for desispec

Edmond

sbailey:
Looks good, thanks.  Merging now.

PR #195, opened 2021-06-23
fix priors test; add github actions

sbailey:
This PR fixes a unit test that was broken by #194, and adds github actions tests.  We had missed that the unit tests weren't working because travis is not longer running but we hadn't ported to github actions and didn't notice that the tests weren't running at all.

Tests pass at NERSC, but I expect this will take a few iterations to get the github actions config sorted out so apologies in advance for the extra emails...


sbailey:
@weaverba137 could you please work your magic to add `secrets.COVERALLS_REPO_TOKEN` to enable coverage tests from GitHub actions on this repo?  Thanks.

weaverba137:
You should be good to go.

sbailey:
Thanks @weaverba137, merging now.

For the record, the other change that slipped in was to support the latest version of h5py which has removed support for `dataset.value` syntax, requiring now the more cryptic `dataset[()]` to access the value of a scalar.  This increased my interest in dropping h5py from redrock.

PR #199, opened 2021-10-21
update top hat prior

echaussidon:
Hi @sbailey, @akremin,

Here the update of the prior function to avoid fail when the redshift prior is close to 0.05. 

I test it in the login node of nersc and it should work. 

Edmond

### DIFF
--- a/py/redrock/priors.py
+++ b/py/redrock/priors.py
@@ -101,11 +101,11 @@ class Priors():
         Returns:
             prior values on the redshift grid.
         """
-
-        tophat_in_xi2_unit = np.vectorize(lambda x : 0 if np.abs(x - z0) < s0/2 else np.NaN)
-
-        prior = tophat_in_xi2_unit(z)
+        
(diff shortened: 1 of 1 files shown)


echaussidon:
@sbailey, I dont' understand the error here .. 

Do you plan to merge this ? 

Edmond

sbailey:
Thanks, @echaussidon.  I had lost track of this one.  The test failures were unrelated to your PR.  I'll merge now, but could you explain a little more what the failure was and how this fixes it?  The new code looks cleaner but I don't see what is actually different.  e.g. what values of s0, z0, and z would make it succeed/fail?  Thanks.

echaussidon:
It was not clear for me but here what I understood:

The problem came from np.vectorize which failed at some point when the redshift prior is close to 0.05. I don't know why ... but I isolated the problem and I juste realized that I didn't need to use np.vectorize so I removed it.

PR #202, opened 2022-01-23
propagate specgroup related keywords from input

sbailey:
This is a companion PR to desihub/desispec#1618 to propagate spectral grouping related keywords to the redrock output.  See desispec#1618 for more details about the keywords and motivation.

Note: In principle redrock can be run on more than one input file and this PR only propagates keywords from the first input file, without checking whether other files are consistent.  This is the minimal change needed for Fuji which only runs on a single file, but might bite us in the future.

@moustakas


moustakas (approved):
Please merge when ready. Oh and we should update the data model at some point...

Issue #191, opened 2021-03-25
rrdesi_mpi performance notes 


dmargala:
This is a summary of quick performance study of `rrdesi_mpi`. From the profile (see image below), it looks like the computation bottleneck is in `calc_zchi2`/`calc_zchi2_one` at the dot product between a sparse resolution matrix and a spectral template [here](https://github.com/desihub/redrock/blob/4097e1e68449414eb17918cc918c0db8d2db81a3/py/redrock/zscan.py#L84). For GPU-offloading, it might be beneficial to "stack" or "batch" the operations in `calc_zchi2`/`calc_zchi2_one` but it would require deeper investigation in order to understand how targets/templates/redshifts are distributed amongst mpi ranks. 

Single Cori Haswell node performance:

```
#- DESI environment used for cascades, swapping out your own copy of redrock
source /global/cfs/cdirs/desi/software/desi_environment.sh 21.2
module unload redrock

git clone https://github.com/desihub/redrock
export PYTHONPATH=$(pwd)/redrock/py:$PYTHONPATH
export PATH=$(pwd)/redrock/bin:$PATH
export RR_TEMPLATE_DIR=/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/redrock-templates/0.7.2

#- Run redrock with 32x MPI parallelism on an interactive node (~3m20s)
salloc -N 1 -C haswell -q interactive -t 1:00:00
export OMP_NUM_THREADS=1
cd /global/cfs/cdirs/desi/spectro/redux/cascades/tiles/80605/20201215/
time srun -n 32 -c 2 rrdesi_mpi spectra-0-80605-20201215.fits -o $SCRATCH/redrock-0-80605-20201215.h5 -z $SCRATCH/zbest-0-80605-20201215.fits
...
Computing redshifts took: 147.6 seconds
Writing zscan data took: 1.1 seconds
Writing zbest data took: 24.9 seconds
Total run time: 191.2 seconds

real	3m18.792s
user	0m0.071s
sys	0m0.037s
```

Use `python -m cProfile ... ` with a single mpi rank to generate a profile:

```
srun -n 1 -c 2 --cpu-bind=cores python -m cProfile -o profile.pstats $(which rrdesi_mpi) spectra-0-80605-20201215.fits -o $SCRATCH/redrock-0-80605-20201215.h5 -z $SCRATCH/zbest-0-80605-20201215.fits
...
Computing redshifts took: 2639.9 seconds
[...]

PR #204, opened 2022-02-17
gpu-ify zscan

dmargala:
This PR adds the ability to use GPUs during the "zscan" step of redrock (see #191 for cpu performance analysis). The GPU-ification is implemented using CuPy. 

A few changes were made to expose larger chunks of data to the processes that use GPUs for the zscan step:
 * The `ReDistTemplate` class takes advantage of the distributed initialization implemented in `DistTemplate` but "allgathers" the rebinned template pieces so each process has the full redshift range of the template.
 * Add an option to `DistTargetsDESI` to distribute the targets according to a processes "capacity". On Perlmutter GPU node, it seems like the GPU processes can handle about 15 times as many targets at the CPU processes.
 * Since the "zfind" step has not been gpu-ifyed yet, the lopsided target distribution needs to be rebalanced after zscan.

GPU usage is enabled in `rrdesi_mpi` by specifying the `--gpu` flag. There is an additional option (`--max-gpuprocs`) to limit number of MPI processes using GPUs. GPU utilization during "zscan" is pretty well saturation with 1 MPI process per GPU but it's useful to have more MPI processes for the subsequent "fitz" step.

```
salloc -C gpu -N 1 -G 4 -t 30 -q interactive -A desi_g

#- load environment
module load cudatoolkit
module load python
source activate redrock-p8r
export RR_TEMPLATE_DIR=/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/redrock-templates/0.7.2
export OMP_NUM_THREADS=1
export MPICH_GPU_SUPPORT_ENABLED=1

#- copy a redrock coadd input file to the current directory for testing
cp /global/cfs/cdirs/desi/spectro/redux/everest/tiles/cumulative/80618/20210328/coadd-3-80618-thru20210328.fits .

#- example existing usage
srun -n 64 -c 2 --cpu-bind=cores rrdesi_mpi -i coadd-3-80618-thru20210328.fits -o redrock-3-80618-thru20210328.fits -d rrdetails-3-80618-thru20210328.h5

#- new command to use gpus
srun -n 64 -c 2 --cpu-bind=cores --gpu-bind=map_gpu:3,2,1,0 rrdesi_mpi -i coadd-3-80618-thru20210328.fits -o redrock-3-80618-thru20210328.fits -d rrdetails-3-80618-thru20210328.h5 --gpu --max-gpuprocs 4
```

Early performance benchmarks indicate about a ~1.4x speedup relative to the CPU version on a Perlmutter GPU node. Compare the GPU_V5 (brown) with CPU baseline (blue) in the following figure of the walltime vs mpi ranks for various iterations of the GPU implementation. 

image

According to fitsdiff, there are some relatively minor looking changes to the output data using gpus when compared with the original cpu output.
```
fitsdiff /global/cfs/cdirs/desi/spectro/redux/everest/tiles/cumulative/80618/20210328/redrock-3-80618-thru20210328.fits redrock-3-80618-thru20210328.fits
...
     Column CHI2 data differs in row 0:
        a> 47268.54813293938
         ?                ^^
        b> 47268.5481329394
         ?                ^
     Column CHI2 data differs in row 2:
        a> 8325.458265185356
        b> 8325.458264797926
     Column CHI2 data differs in row 3:
        a> 6647.635786965489
         ?             ^ ^^
        b> 6647.635786995292
         ?             ^ ^ +
     Column CHI2 data differs in row 4:
        a> 21779.621988100713
         ?                  ^
        b> 21779.621988100716
         ?                  ^
     Column CHI2 data differs in row 5:
        a> 8419.153264559805
         ?             ^ - ^
        b> 8419.153264544904
         ?             ^^  ^
     Column CHI2 data differs in row 6:
        a> 43655.38019606759
         ?                 ^
        b> 43655.380196067556
         ?                 ^^
     Column CHI2 data differs in row 8:
        a> 10499.721163630486
         ?             ^^ - ^
        b> 10499.72116357088
         ?             ^^  ^
     Column CHI2 data differs in row 9:
        a> 8836.548374652863
         ?              ----
        b> 8836.548374682665
         ?             ++++
     Column CHI2 data differs in row 10:
        a> 8529.76254041493
         ?             ^^
        b> 8529.76254042983
         ?             ^ +
     Column CHI2 data differs in row 11:
        a> 102536.45528930357
         ?                  -
        b> 102536.4552893035
     ...2404 additional difference(s) found.
     ...
     2414 different table data element(s) found (43.89% different).
```
@marcelo-alvarez @sbailey


py/redrock/utils.py
```diff
@@ -284,6 +284,51 @@ def distribute_partition(A, k):
     return dist
 
 
+def distribute_work_lopsided(nproc, ids, weights=None, capacities=None):
```
marcelo-alvarez, line 287: @dmargala does `distribute_work_lopsided` behave effectively the same as `distribute_work` when the `capacities` argument is not provided? If this is the case, I don't see why `distribute_work` shouldn't be replaced by `distribute_work_lopsided` in order to simplify maintenance.

dmargala: In that case, they should behave effectively the same although the ordering of partitioned task ids is different. The implementation in `distribute_work` looks pretty fancy and I wasn't sure how to modify it to incorporate per-process `capacities`. It's probably more efficient than my lazy `O(n^2)` implementation in `distribute_work_lopsided`. 

```python
In [1]: from redrock.utils import distribute_work, distribute_work_lopsided

In [2]: nprocs, ntasks = 4, 16

In [3]: ids = range(ntasks)

In [4]: distribute_work(nprocs, ids)
Out[4]: [[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11], [12, 13, 14, 15]]

In [5]: distribute_work_lopsided(nprocs, ids)
Out[5]: [[0, 4, 8, 12], [1, 5, 9, 13], [2, 6, 10, 14], [3, 7, 11, 15]]
```

Adding in weights:

```python
In [6]: weights = dict(zip(ids, [1] * ntasks))

In [7]: weights[3] = 5

In [8]: distribute_work(nprocs, ids, weights)
Out[8]: [[0, 1, 2], [3, 4, 5, 6], [7, 8, 9, 10, 11], [12, 13, 14, 15]]

In [9]: distribute_work_lopsided(nprocs, ids, weights)
Out[9]: [[3], [0, 4, 7, 10, 13], [1, 5, 8, 11, 14], [2, 6, 9, 12, 15]]
```

I'm not sure what `distribute_work` is doing there. Total weight per process using `distribute_work` is `3, 8, 5, 3`. Total weight per process using `distribute_work_lopsided` is `5, 5, 5, 5`.

Perhaps that should be spun off into a separate issue?

(while looking into this, I noticed that the weighted task sort in `distribute_work_lopsided` should be reversed  and pushed an update).

sbailey:
It looks like the original distribute_work is implementing the "painter's partition problem" which has the additional constraint that the subgroups must be contiguous ranges of the original input set.  For a completely embarrassingly parallel problem that is an unnecessary constraint that results in less optimal balancing.

However, there might be something fancy going on with the MPI implementation of redrock and how it distributes data and work across multiple nodes, e.g. trying to make sure that a rank is assigned to targets already in memory on that node and not shuffling data across nodes more than necessary.  Maybe.  I haven't studied in detail how it does its data shufflling, but check that a bit and do some profiling when running on multiple nodes.

But +1 for the general concept of replacing the two functions with one that can do weighted or unweighted partitioning.

FWIW, [`desispec.parallel.weighted_partition`](https://github.com/desihub/desispec/blob/bfda8e5cb56c83a617206fca28a80d002077d1ea/py/desispec/parallel.py#L269), and `parallel.dist_*` are similar implementations of the same problem (but we don't want core redrock code to depend upon desispec, to keep it more generally useful).

py/redrock/utils.py
marcelo-alvarez: @dmargala  I'm trying to understand how important `distribute_work_lopsided` is in the speedups you have reported. 

Above you mentioned that the "zfind" step is not gpu-ifyed while "zscan" is, so the allocation of processes using `distribute_work_lopsided` needs to be done separately for "zfind" and "zscan". This implies that "zfind" tasks depend on all "zscan" tasks to complete before running, but that there are situations in which some processes perform work for the "zscan" step on CPU (e.g. capacity=1) and others on GPU processes (e.g. capacity=15). Was this the case in your profiling? Naively I would expect only GPU processes to be working on tasks that have been optimized for GPU would be near optimal, with the gain from using CPUs on those tasks at the same time to be negligible.

dmargala:
Thanks for the background Stephen. 

I still think something is off with the weighted `distribute_work` function. By hand, I would come up with `[[0, 1, 2], [3], [4, 5, 6, 7, 8, 9], [10, 11, 12, 13, 14, 15]` resulting in `3, 5, 6, 6` for my earlier example with weights.

py/redrock/utils.py
marcelo-alvarez: That's helpful indeed. Comparing V3 to V4 at numtasks = 4 to 16, it looks like adding the lopsided make the performance slightly worse. But it seems like there is no scaling test in these plot that shows whether lopsided by itself helps post-GPU_V5. I think it's fine keeping separate functions as is for now and coming back to this after testing on nightly processing and/or we have a better idea of the actual algorithmic differences between the two functions. Thanks.

marcelo-alvarez:
The expected behavior of `distribute_work` is obtained by replacing [lines 247-248 in utils.py](https://github.com/desihub/redrock/blob/master/py/redrock/utils.py#L247-L248):
```python 
            over = curweight - goal
            curweight = elwts + over
```
with
```python
            curweight = elwts
```
This modification ensures that the partitioning encoded in `max_per_proc` limits the size of subgroups.  Running with this modification we obtain
```python
In [6]: weights = dict(zip(ids, [1] * ntasks))
In [7]: weights[3] = 5
In [8]: distribute_work(nprocs, ids, weights)
Out[8]:[[0, 1, 2], [3, 4], [5, 6, 7, 8, 9, 10], [11, 12, 13, 14, 15]]
```
which is one of the optimal results, with a load per process of `3, 6, 6, 5`.  

It is unclear to me why the `goal` variable (the mean amount of work per process) is used in determining whether the amount of work in a given subgroup has exceeded the partition size. In the particular example here, the value of `over` was negative (the first group gets only 3 work units while the average is 5), so the second group (8 work units) ended up being 2 work units above the maximum (6 work units) set by the partitioning.

I will spin this off into a separate issue.

marcelo-alvarez:
> 
> ```
> salloc -C gpu -N 1 -G 4 -t 30 -q interactive -A desi_g
> 
> #- load environment
> module load cudatoolkit
> module load python
> source activate redrock-p8r
> export RR_TEMPLATE_DIR=/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/redrock-templates/0.7.2
> export OMP_NUM_THREADS=1
> export MPICH_GPU_SUPPORT_ENABLED=1
> 
> #- copy a redrock coadd input file to the current directory for testing
> cp /global/cfs/cdirs/desi/spectro/redux/everest/tiles/cumulative/80618/20210328/coadd-3-80618-thru20210328.fits .
> 
> #- example existing usage
> srun -n 64 -c 2 --cpu-bind=cores rrdesi_mpi -i coadd-3-80618-thru20210328.fits -o redrock-3-80618-thru20210328.fits -d rrdetails-3-80618-thru20210328.h5
> 
> #- new command to use gpus
> srun -n 64 -c 2 --cpu-bind=cores --gpu-bind=map_gpu:3,2,1,0 rrdesi_mpi -i coadd-3-80618-thru20210328.fits -o redrock-3-80618-thru20210328.fits -d rrdetails-3-80618-thru20210328.h5 --gpu --max-gpuprocs 4
> ```
> 
> Early performance benchmarks indicate about a ~1.4x speedup relative to the CPU version on a Perlmutter GPU node. Compare the GPU_V5 (brown) with CPU baseline (blue) in the following figure of the walltime vs mpi ranks for various iterations of the GPU implementation.

I have run the test above under the "22.1b" tagged environment on Perlmutter, i.e. with 
```
module load cudatoolkit
module load python
source activate redrock-p8r
```
replaced by
```
source /global/common/software/desi/desi_environment.sh 22.1b
```
and 
```
git clone -b gpu-zscan https://github.com/desihub/redrock 
export PYTHONPATH=$PWD/redrock/py:$PYTHONPATH
export PATH=$PWD/redrock/bin:$PATH
```
prior to running the test for this branch, so that `redrock/0.15.2` is compared to this branch with and without GPU acceleration enabled. 

This test can be run with the command
```
/global/cfs/cdirs/desi/users/malvarez/run-gpuify-zscan-test.sh
```
on Perlmutter. I obtain the following results:
```
Test results:
  walltime 0.15.2:           68.601
  walltime gpu-zscan (gpu):  62.773
  walltime gpu-zscan (cpu):  68.907

  reported 0.15.2:           51.7
  reported gpu-zscan (gpu):  45.3
  reported gpu-zscan (cpu):  51.9
```
I find a speedup of a factor of only 1.1 in wall time, or about 1.15 in the time reported by the code itself. @dmargala is this the expected speedup for this command? It looks to me low compared to the plot you attached above for numtasks=64. Thanks.

marcelo-alvarez:
No differences in output or performance between this branch in CPU mode and master, on either Cori Haswell and Perlmutter GPU partitions. Additionally, some speed up is observed when following the instructions given above. Merging this  now. Thanks @dmargala for these changes.

PR #207, opened 2022-02-19
Simplify distribution of work to processes 

marcelo-alvarez:
This modification addresses an issue that was discovered with @dmargala during review for #204, specifically undocumented features in the `distribute_work` function.


sbailey (approved):
Looks good.  I independently verified that it produces bitwise identical results in the same runtime with simpler code.  Sounds good to me...

PR #211, opened 2022-04-28
Created new GPU kernels for zscan that are called by new method

craigwarner-ufastro:
calc_zchi2_gpu_new.  The new kernels result in all templates being
processed in parallel for each target.

There are 4 raw CUDA kernels loaded by CUPY and called from calc_zchi2_gpu_new:
batch_dot_product
calc_M_and_y_atomic
tb_zc_dot
calc_z_prod

Additionally cupy.linalg.solve replaces numpy.linalg.solve to solve for
all templates for one object simultaneously.

When running in GPU mode, the new version also allocates 0 targets to
CPU procs that are not GPU enabled.


sbailey:
This is an impressive speedup.  I independently verified some timing numbers for 500 targets, which are consistent with numbers previously reported offlist, but I'll put them here for the record:

| System | Scanning redshifts [sec] | Total runtime [sec] (except srun/import startup overhead |
| ------- | ------------------------ | --------------------------------------------------- |
| Cori Haswell 32-core                | 83 | 114 |
| Perlmutter P1 64-core              | 30 | 51 |
| Perlmutter P1 4 GPU - master | 18 | 45 | 
| Perlmutter P1 4 GPU - this PR | 5 | 33 |

The cuda kernels are clearly documented and I especially appreciate the comments like
```
// This replicates the python commands:
//     M = Tb.T.dot(np.multiply(weights[:,None], Tb))
//     y = Tb.T.dot(wflux)
```
to help provide context for what the kernel is doing.

However, I think this takes us in the wrong direction for maintainability and flexibility, including introducing a new language into the DESI stack (cuda C).  If the algorithm was completely frozen and we were only making it faster, this approach could be ok.  In practice we are still exploring ideas like using more expensive fitting algorithms (like non-negative fits) and/or augmenting the templates with per-camera observer frame offsets simultaneously fit with the templates themselves.  This PR has absorbed enough algorithmic details into the cuda C kernels that I would find it hard to explore these ideas while maintaining CPU/python and GPU/python/cuda consistency.

I think we need to revisit the structure of the data flow and try to make the CPU and GPU paths more similar, more aggressively leveraging the API-symmetry of cupy and numpy.  We should resort to pure cuda kernels only if absolutely necessary for a big win, and even then try to maintain a 1:1 mapping between the cuda kernal and an equivalent CPU function.

`calc_zchi` and `calc_zchi_gpu` already have ~50% duplicated code; and there is further conceptual duplication in `calc_zchi2_one`, `_zchi2_one`, and `_zchi2_batch` that could likely be unified.  Let's take a step back and try to streamline this for both CPU and GPU, while keeping separate concepts like:
  * redshifting + rebinning the templates to the observed wavelength grid
  * multiplying the redshifted+rebinned templates by the per-target per-spectrum resolution matrices
  * fitting those resolution-convolved redshifted+rebinned templates to data (currently as linear algebra, but in the future could involve non-negativity constraints or the addition of nuisance terms like offsets).

Let's put this branch on the back burner and try another approach to see if we can get similar speedups while improving the flexibility and maintainability of the current code.  I realize that's vague, but I don't think we should merge this branch as-is, despite the significant speedups.

craigwarner-ufastro:
This draft PR is intended for a code review and has multiple options for algorithms, only one of which will remain in the final version.  The code is documented with extra comments for this purpose that start with "###!!! NOTE - comment".

The original main method for redshift scanning is _calc_zchi2()_ and the original GPU method is _calc_zchi2_gpu()_.

My first attempt at accelerating the code (version 1) is _calc_zchi2_gpu_new()_ which was intended as a drop-in replacement for calc_zchi2_gpu.  It is the fastest version, with a runtime of 10.4s on 1 GPU / 1 CPU and 4.0s on 4 GPUs compared to 29.7s for the original CPU version on 64 cores and 84.3s for the original GPU version on 1 CPU / 1GPU and 22.6s for 4 GPU / 4 CPU.

There were some concerns with the first attempt being less maintainable though because of customized CUDA kernels that replaced multiple operations at once so Stephen asked me to try to redesign to be more maintainable, including reorganizing the CPU code so that everything had the same data flow.  Also an attempt was made to use pure CUPY operations when possible instead of custom kernels.  For some operations this is not possible though, e.g. the lion's share of my speed-up is due to my _batch_dot_product_sparse_ kernel, which computes a dot product between the sparse matrix for each spectrum and the batch of templates.  What I was able to do though is restrict custom kernels to a basic operation such as batch_dot_product and remove the multi-operation kernels, leaving us with only 2 custom kernels that are both batch dot products for different array types.

My attempt at doing this is the version 2 algorithm, _calc_zchi2_v2()_ in which the data flow is like:
`if gpu:
    do A in batch
else:
    for i in nz:
        do A
        
if gpu:
    do B in batch
else:
    for i in nz:
        do B`
The pros of version 2 include that it follows the same data flow either on the CPU or GPU by doing every step in batch for every template before proceeding to the next operation and that it maintains the lion's share of the speed gains of version 1, running in 13.6s on 1CPU / 1 GPU and 4.8s on 4 CPU /  4 GPU.

The con is that for reasons I don't understand that may have to do with shared memory and MPI, it is slower than the original CPU version by about 16% on 64 CPUs despite being faster by 11% on 1 CPU.

Given this surprising result, I tried one more version - version 3, which is _calc_zchi2_batch_v3()_.
**This is the version I would advocate we use.**
Version 3 doesn't quite maintain the same data flow in that it maintains the outer loop for the CPU version but the CPU and GPU operations are quite clearly analogous and I believe easily maintainable, e.g.
`if gpu:
    do A in batch
    do B in batch
else:
    for i in nz:
        do A
        do B`

Finally, I discovered one more CUPY inefficiency that seems to appear only on Volta architectures like my older GeForce 1660 Super consumer grade GPU - presumably the inefficiency would still be there on V100 data center GPUs though.  It is not present on the A100 Ampere architecture.

When executing the code
`all_M = Tbs.swapaxes(-2, -1) @ (weights[None, :, None] * Tbs)
all_y = (Tbs.swapaxes(-2, -1) @ wflux)`
on my GeForce for --ntarget 10, I get 18.8 seconds for scanning redshifts and when I profile it, 16.6 seconds are spent using cudaMemcpy and 16.3 seconds with something called _volta_dgemm_64x64_nn_.  By comparison when I run my custom kernel to calculate M and y from version 1 of my code, I get 2.7 seconds total runtime, with 1.2s spent in cudaMemcpy and 734ms for my custom kernel to run.

Meanwhile on Perlmutter, the difference in time for running all targets is 13.6s vs 10.6s, and there is no large increase in the time spent in cudaMemcpy.

Thus I designed one more custom kernel, again to do only the dot product (and transpose) parts of calculating the M arrays, _batch_dot_product_3d3d_gpu()_, which results in runtimes on my machine of 3.6s for --ntarget 10 and on Perlmutter of 13.1s for all targets - vastly improved for Volta architectures and slightly improved for Ampere.  I keep the calculation of the y array as pure CUPY and only the dot product portion of the M calculation is thus offloaded to a custom kernel.

craigwarner-ufastro:
Here is a spreadsheet of various timings of version 0 (original code) versus versions 1, 2, and 3. I time 5 individual algorithms plus the overall time.  Time No Sync is the actual run time for the GPU versions as to get the timings for each algorithm I have do a CUPY sync after each one, which slows things down slightly (e.g. I ran once with sync after each step to get the timings of each step and then a second time with the syncs removed to just get the actual overall run time).Note that I created this spreadsheet before realizing the CUPY matrix multiplication issue in calculating M that I describe above so this is for the pure CUPY computation of M and y -- the version that I actually pushed is 0.5s faster on 1 GPU and 0.1s faster on 4 GPUs.

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head>
	
	
	
</head>

<body>


Version | GPU | CPU | Total time | T0 | T1 | T2 | T3 | T4 | Other | Time No Sync
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
0 | 0 | 1 | 1074.9 | 658.6 | 332.8 | 15.8 | 30.4 | 17.6 | 19.7 |  
2 | 0 | 1 | 951.3 | 584.3 | 299.2 | 9.4 | 40.6 | 15.2 | 2.6 |  
3 | 0 | 1 | 1009.0 | 645.0 | 296.3 | 14.8 | 31.0 | 17.3 | 4.6 |  
  |   |   |   |   |   |   |   |   |   |  
0 | 0 | 64 | 29.7 | 16.8 | 7.5 | 0.5 | 0.8 | 0.5 | 3.7 |  
2 | 0 | 64 | 34.8 | 16.3 | 9.2 | 0.5 | 2.5 | 0.6 | 5.7 |  
3 | 0 | 64 | 28.8 | 16.5 | 7.3 | 0.5 | 0.8 | 0.5 | 3.2 |  
  |   |   |   |   |   |   |   |   |   |  
0 | 1 | 1 | 84.3 | 62.5 | 15.2 | 0.8 | 3.4 | 0.7 | 1.7 | 82.1
1 | 1 | 1 | 11.2 | 4.4 | 2.9 | 1.1 | 0.7 | 0.6 | 1.5 | 10.4
2 | 1 | 1 | 14.6 | 4.5 | 6.2 | 0.7 | 0.8 | 1.0 | 1.4 | 13.6
3 | 1 | 1 | 14.6 | 4.5 | 6.2 | 0.7 | 0.8 | 1.0 | 1.4 | 13.6
  |   |   |   |   |   |   |   |   |   |  
0 | 4 | 4 | 22.6 | 16.0 | 3.8 | 0.2 | 0.9 | 0.2 | 1.5 | 22.1
1 | 4 | 4 | 4.2 | 1.1 | 0.7 | 0.6 | 0.2 | 0.2 | 1.4 | 4.0
2 | 4 | 4 | 5.1 | 1.2 | 2.0 | 0.2 | 0.2 | 0.2 | 1.3 | 4.8
3 | 4 | 4 | 5.0 | 1.2 | 1.9 | 0.2 | 0.2 | 0.2 | 1.2 | 4.8
  |   |   |   |   |   |   |   |   |   |  
0 | 4 | 64 | 16.9 | 9.5 | 1.8 | 0.1 | 0.4 | 0.1 | 5.0 | 16.6
1 | 4 | 64 | 4.4 | 1.2 | 0.7 | 0.8 | 0.2 | 0.2 | 1.3 | 4.2
2 | 4 | 64 | 5.4 | 1.2 | 2.0 | 0.2 | 0.2 | 0.3 | 1.5 | 5.0
3 | 4 | 64 | 5.3 | 1.2 | 2.0 | 0.2 | 0.2 | 0.3 | 1.4 | 5.0
  |   |   |   |   |   |   |   |   |   |  
T0 | Batch dot product of sparse matrices of spectra with all templates |   |   |   |   |   |   |   |   |  
T1 | Calculation of M and y arrays |   |   |   |   |   |   |   |   |  
T2 | Use linalg.solve to solve M x zcoeff = y for zcoeff |   |   |   |   |   |   |   |   |  
T3 | Batch dot product of Tbs array and zcoeff to calculate model |   |   |   |   |   |   |   |   |  
T4 | Final zchi2 calculation from model, flux, and weights |   |   |   |   |   |   |   |   |  
Other | Total time minus the above |   |   |   |   |   |   |   |   |  


</body>

</html>

sbailey:
Trying the new `rebin.trapz_rebin_batch_gpu` crashes in a way that I don't know how to debug, which highlights my concern about introducing custom cuda kernels instead of using higher-level albeit slower methods like cupy:
```python
import numpy as np
from redrock import rebin
x = np.arange(10)
y = np.arange(10)
rebin.trapz_rebin(x, y, edges=[0, 2, 4, 5, 6])
rebin.trapz_rebin_batch_gpu(x, y, edges=[0, 2, 4, 5, 6])
```
results in
```
Out[6]: ---------------------------------------------------------------------------
CUDARuntimeError                          Traceback (most recent call last)
File /global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/conda/lib/python3.9/site-packages/IPython/core/formatters.py:707, in PlainTextFormatter.__call__(self, obj)
    700 stream = StringIO()
    701 printer = pretty.RepresentationPrinter(stream, self.verbose,
    702     self.max_width, self.newline,
    703     max_seq_length=self.max_seq_length,
    704     singleton_pprinters=self.singleton_printers,
    705     type_pprinters=self.type_printers,
    706     deferred_pprinters=self.deferred_printers)
--> 707 printer.pretty(obj)
    708 printer.flush()
    709 return stream.getvalue()

File /global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/conda/lib/python3.9/site-packages/IPython/lib/pretty.py:410, in RepresentationPrinter.pretty(self, obj)
    407                         return meth(obj, self, cycle)
    408                 if cls is not object \
    409                         and callable(cls.__dict__.get('__repr__')):
--> 410                     return _repr_pprint(obj, self, cycle)
    412     return _default_pprint(obj, self, cycle)
    413 finally:

File /global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/conda/lib/python3.9/site-packages/IPython/lib/pretty.py:778, in _repr_pprint(obj, p, cycle)
    776 """A pprint that just redirects to the normal repr function."""
    777 # Find newlines and replace them with p.break_()
--> 778 output = repr(obj)
    779 lines = output.splitlines()
    780 with p.group():

File cupy/_core/core.pyx:1621, in cupy._core.core.ndarray.__repr__()

File cupy/_core/core.pyx:1735, in cupy._core.core.ndarray.get()

File cupy/cuda/memory.pyx:533, in cupy.cuda.memory.MemoryPointer.copy_to_host()

File cupy_backends/cuda/api/runtime.pyx:552, in cupy_backends.cuda.api.runtime.memcpy()

File cupy_backends/cuda/api/runtime.pyx:132, in cupy_backends.cuda.api.runtime.check_status()

CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered
```
This could easily be a GPU-newbie user error: I'm trying to access a GPU function from ipython on a perlmutter login node.  I think that is supposed to work, but are there other environment variables, call prefixes, etc. that I need to be using to actually access the GPU?

py/redrock/utils.py
```diff
@@ -216,26 +216,82 @@ def distribute_work(nproc, ids, weights=None, capacities=None):
...
+    This returns the transmitted flux fraction:
+    1 -> everything is transmitted (medium is transparent)
+    0 -> nothing is transmitted (medium is opaque)
+    Args:
+        all_zObj (array of float): Redshift of all objects
+        lObs (array of float): wavelength grid
+    Returns:
+        cp.array of float (nz x nlambda): transmitted flux fraction
+    """
+    import cupy as cp
```
sbailey, line 279: There is a lot of replicated code in `transmission_Lyman_gpu` compared to `transmission_Lyman`, differing only by `np` vs. `cp`.  Let's use the API similarity between those two libraries to reduce this to a single function that automatically does the right thing with either np or cp arrays.  Similarly the scalar vs. array `zObj` can be handled with `np.atleast_1d` to promote it to an array, do the calculation, and then special case whether the return value should be sliced back to a 1D array or left as a 2D array.

py/redrock/rebin.py
```diff
@@ -10,6 +10,169 @@
...
+                    //- Otherwise the samples span over this bin
+                    ylo = y[y_idx] + (edges[ibin]-xj) * (y[y_idx] - y[y_idx-1]) / (xj - xj1);
+                    yhi = y[y_idx] + (edges[ibin+1]-xj) * (y[y_idx] - y[y_idx-1]) / (xj - xj1);
+                    area = 0.5 * (ylo+yhi) * (edges[ibin+1]-edges[ibin]);
+                    r += area;
+                }
+                result[i] = r / (edges[ibin+1]-edges[ibin]);
+            }
+
+            __global__ void batch_trapz_rebin_uneven(const double* x, const double* y, const double* edges, const double* myz, const int* idx, double* result, int nz, int nbin, int nbasis, int nt) {
```
sbailey, line 101: Evenly spaced bins is a special case of unevenly spaced bins.  Is the performance improvement from "batch_trapz_rebin" enough better than "batch_trapz_rebin_uneven" to be worth maintaining two separate kernels in addition to their python wrappers and the non-GPU numba version?

py/redrock/rebin.py
```diff
@@ -10,6 +10,169 @@
...
+cp = None
+
+#Since CUPY import is moved, vars block_size and cuda_source will always be
+#set but this should cost no real time.
+block_size = 512 #Default block size, should work on all modern nVidia GPUs
+
+# cuda_source contains raw CUDA kernels to be loaded as CUPY module
+cuda_source = r'''
+        extern "C" {
+            __global__ void batch_trapz_rebin(const double* x, const double* y, const double* edges, const double* myz, double* result, int nz, int nbin, int nbasis, int nt) {
```
sbailey, line 23: Did you try implementing these with cupy or [numba for cuda](https://numba.readthedocs.io/en/stable/cuda/overview.html#overview) first?  Very few people within the collaboration know pure-cuda programming, so for maintainability it would be highly preferable to keep these in a higher-level more python-like form, and only resort to pure cuda kernels when absolutely necessary to speed up a function whose slowness is a blocking factor and isn't amenable to other GPUification.

py/redrock/test/test_rebin.py
```diff
@@ -74,6 +85,40 @@ def test_nonuniform(self):
         self.assertTrue(np.allclose(yy[0:2], 2/np.pi, atol=5e-4))
         self.assertTrue(np.allclose(yy[2:4], -2/np.pi, atol=5e-4))
 
+    def test_gpu_trpazrebin(self):
```
sbailey, line 88: minor: trpa typo :)

py/redrock/test/test_rebin.py
```diff
@@ -74,6 +85,40 @@ def test_nonuniform(self):
...
+        '''test rebinning unevenly spaced x for GPU vs CPU'''
+        if (not cp_available):
+            self.assertTrue(True)
+            return
+        x = np.linspace(0, 10, 100)**2
+        y = np.sqrt(x)
+        edges = np.linspace(0,100,21)
+        c = rebin.trapz_rebin(x, y, edges=edges)
+        g = rebin.trapz_rebin_batch_gpu(x, y, edges=edges)
+        self.assertTrue(np.allclose(c,g))
```
sbailey, line 120: Thanks for including these tests.  We should get into the habit of writing comparison tests like this any time we have alternate code paths for the same algorithm.

py/redrock/utils.py
```diff
@@ -216,22 +216,51 @@ def distribute_work(nproc, ids, weights=None, capacities=None):
...
     0 -> nothing is transmitted (medium is opaque)
+
+    This method will handle 3 options:
+    1 -> For GPU mode, zObj is an array of all z for a template and the return
+    value will be a cupy array (nz x nlambda)
+    2-> In CPU mode, it can auto-detect if zObj is a numpy array and if so,
+    again, all z will be processed as a vector and the return value
+    will be a numpy array (nz x nlambda)
+    3-> For legacy, it is still supported to pass zObj as a float and
+    in this case, the return value will be a 1-d numpy array (nlambda).
```
sbailey, line 232: Do the unit tests cover these 3 cases, including that the return type is as expected?  cupy vs. numpy array propagation is still a bit of a mystery to me (e.g. does `T = np.ones_like(lRF)` return a cupy array if `lRF` is a cupy array?) so it could be helpful to explicitly confirm that it does what we think and protect against future updates that might accidentally trigger a GPU -> CPU array transition.

sbailey (commented):
I'm submitting another set of partial review comments about the rebinning to make sure they don't get lost before I have time to continue reviewing the zscan portion.

py/redrock/rebin.py
```diff
@@ -125,24 +288,190 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+        >>> x = np.arange(10)
+        >>> y = np.ones(10)
+        >>> trapz_rebin(x, y, edges=[0,2,4,6,8])  #- density still 1, not 2
+        array([ 1.,  1.,  1.,  1.])
+
+    Args:
+        x (array): input x values.
+        y (array): input y values (for all bases).
+        edges (array): (optional) new bin edges.
+        myz (array): redshifts
```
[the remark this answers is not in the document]
craigwarner-ufastro: 1. Ok - `x` and `edges` are 1-d just as in the CPU method.  `y` may be 1-d or 2-d.  I will update this.

2. Ok - will clarify - one major difference between CPU and GPU version is that on the GPU, multiplying `x` by `(1+z)` is done in the kernel so that it can be done in parallel.

3. Ok - base case is still valid but will add example of 2-d `y` array too.

py/redrock/rebin.py
```diff
@@ -125,24 +288,190 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+    #Launch kernel and syncrhronize
+    batch_trapz_rebin_kernel((blocks,), (block_size,), (x, y, edges, myz, result, nz, nbin, nbasis, nt))
+    #cp.cuda.Stream.null.synchronize()
+
+    #Squeeze array and remove dims of 1
+    if (nz == 1 or nbasis == 1):
+        result = result.squeeze()
+    return result
+
+def trapz_rebin_batch_gpu_unevenly_spaced(x, y, edges, myz):
```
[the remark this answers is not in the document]
craigwarner-ufastro: This issue will go away since it looks like we can just use `trapz_rebin_batch_gpu_unevenly_spaced` without losing speed.  I had developed the other method first before I realized that the QSO templates were unevenly spaced in wavelength.  This will simplify some things.

- For the single user-facing method possibility, would it be ok to pass `myz` with default arg None to that method, slightly changing its existing signature?  If so then I think it is viable but would probably still need to propagate to the current method (branching to that instead of the numba method).

py/redrock/rebin.py
```diff
@@ -125,24 +288,190 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+            where nbin = len(results) = len(edges)-1
+
+    Raises:
+        ValueError: if edges are outside the range of x or if len(x) != len(y)
+
+    """
+    #Use global to import cupy here - it will not be needed to be imported
+    #in any other method
+    global cp
+    import cupy as cp
```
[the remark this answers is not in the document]
craigwarner-ufastro: So the logic here is that using the `global cp`, I didn't need to add an import to every method, just the entry point method.  This is borrowed from what I did in `zscan.py`.  Although here it ceases to be an issue when I merge the two GPU rebin methods so I will drop the `global cp` here.

It is really surprising/annoying that importing cupy is so expensive when running without a GPU - it should be a no-op in that case I would think.

py/redrock/rebin.py
craigwarner-ufastro: I have not done anything with numba for cuda, but it is not possible as far as I can tell in pure-python cupy as its not quite so simple as just performing a couple of vectorized operations.  I can look into numba for cuda but I suspect that will not work either.  However, as this is essentially a low level library operation, I would think the maintenance would be minimal.  I will let you know if after looking into it, I think numba for cuda is an option.

py/redrock/utils.py
craigwarner-ufastro: Yes, the unit tests cover all 3 cases verifying that the results match (or are `allclose`) for all three.  I don't explicitly check the output return type but I can add that easily.  In the method, I explicitly cast `zObj` (and thus `lRF`) as numpy or cupy array depending on the `use_gpu` flag.

And yes, `T = np.ones_like(lRF)` will return a cupy array if `lRF` is a cupy array and a numpy array if it is a numpy array.

sbailey:
`test_gpu_trpazrebin` is failing on a Perlmutter login node:
```
pytest py/redrock
...
=================================== FAILURES ===================================
________________________ TestRebin.test_gpu_trpazrebin _________________________

self = <redrock.test.test_rebin.TestRebin testMethod=test_gpu_trpazrebin>

    def test_gpu_trpazrebin(self):
        '''Test that GPU version matches CPU for constant flux density at various binnings'''
        if (not cp_available):
            self.assertTrue(True)
            return
        nx = 10
        x = np.arange(nx)*1.1
        y = np.ones(nx)
    
        #- test various binnings
        for nedge in range(3,10):
            edges = np.linspace(min(x), max(x), nedge)
            yy = rebin.trapz_rebin_batch_gpu(x, y, edges=edges)
            self.assertTrue(np.all(yy == 1.0), msg=str(yy))
    
        #- edges starting/stopping in the interior
>       summ = rebin.trapz_rebin_batch_gpu(x, y, edges=[0.5, 8.3])[0]

py/redrock/test/test_rebin.py:104: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cupy/_core/core.pyx:1421: in cupy._core.core.ndarray.__getitem__
    ???
cupy/_core/_routines_indexing.pyx:33: in cupy._core._routines_indexing._ndarray_getitem
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed

cupy/_core/_routines_indexing.pyx:365: IndexError
```

py/redrock/utils.py
sbailey: The updated `transmission_Lyman` resolves these concerns, and is a good model for future maintainability of CPU vs. GPU code paths — it's the same code for both with no duplication of core algorithm statements, and very minimal if/else handling of data types for the two cases.

craigwarner-ufastro:
Thank you for catching that - a strong reminder to me to always go back and rerun unit tests after any changes.

I must have added the code to squeeze extraneous dimensions in the result after doing the unit test:
```    #Squeeze array and remove dims of 1
    if (nz == 1 or nbasis == 1):
        result = result.squeeze()
```
and not thought about the case where there is only 1 bin as well, which resulted a numpy array of 0 dimensions but with a scalar value (whereas I would have expected it to reduce to a 1-element, 1-dimensional array).  Will check in the corrected code when I also merge the two trapz_rebin GPU methods.

```    #Squeeze array and remove dims of 1
    if (result.size == 1):
        #Use flatten because squeeze results in array dim 0
        result = result.flatten()
    elif (nz == 1 or nbasis == 1):
        result = result.squeeze()
```

> `test_gpu_trpazrebin` is failing on a Perlmutter login node:
> 
> ```
> pytest py/redrock
> ...
> =================================== FAILURES ===================================
> ________________________ TestRebin.test_gpu_trpazrebin _________________________
> 
> self = <redrock.test.test_rebin.TestRebin testMethod=test_gpu_trpazrebin>
> 
>     def test_gpu_trpazrebin(self):
>         '''Test that GPU version matches CPU for constant flux density at various binnings'''
>         if (not cp_available):
>             self.assertTrue(True)
>             return
>         nx = 10
>         x = np.arange(nx)*1.1
>         y = np.ones(nx)
>     
>         #- test various binnings
>         for nedge in range(3,10):
>             edges = np.linspace(min(x), max(x), nedge)
>             yy = rebin.trapz_rebin_batch_gpu(x, y, edges=edges)
>             self.assertTrue(np.all(yy == 1.0), msg=str(yy))
>     
>         #- edges starting/stopping in the interior
> >       summ = rebin.trapz_rebin_batch_gpu(x, y, edges=[0.5, 8.3])[0]
> 
> py/redrock/test/test_rebin.py:104: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> cupy/_core/core.pyx:1421: in cupy._core.core.ndarray.__getitem__
>     ???
> cupy/_core/_routines_indexing.pyx:33: in cupy._core._routines_indexing._ndarray_getitem
>     ???
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> >   ???
> E   IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
> 
> cupy/_core/_routines_indexing.pyx:365: IndexError
> ```

craigwarner-ufastro:
I just pushed out an update addressing all of the conversations above regarding `rebin.py`, `utils.py`, and their unit tests.

py/redrock/rebin.py
```diff
@@ -125,24 +239,150 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+        result = { hs:np.empty((nz, len(wave), nbasis)) for hs, wave in dwave.items() }
+        for i, z in enumerate(myz):
+            x = (1.+z)*template.wave
+            for hs, wave in dwave.items():
+                for b in range(nbasis):
+                    result[hs][i,:,b] = trapz_rebin(x, template.flux[b], wave)
+    else:
+        #Legacy mode - a single redshift passed - return a dict with rebinned
+        #template for every basis for this one redshift.
+        result = { hs:np.array([ trapz_rebin((1.+myz)*template.wave, template.flux[b], wave) for b in range(nbasis) ]).transpose() for hs, wave in dwave.items() }
```
sbailey, line 386: Currently `rebin_template` with `use_gpu=False` supports scalar or vector `myz`, but `use_gpu=True` only supports vector `myz` and crashes if `myz` is a scalar.  This is another case where were should make the CPU vs. GPU API/feature identical, perhaps my adding support for scalar `myz` directly to `trapz_rebin` (for both `use_gpu=True/False`).

i.e. even if the rest of the GPU code would never call this with a scalar `myz`, for maintainability and PLS(S), every function that supports `use_gpu=True/False` should be API/feature/result identical unless there is a very specific reason not to (e.g. an option like blocksize that only applies to one but not the other).

py/redrock/rebin.py
```diff
@@ -125,24 +239,150 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+        edges = np.asarray(edges)
+    edges = cp.array(edges, dtype=cp.float64)
 
-def rebin_template(template, z, dwave):
+    if (myz is None):
+        myz = cp.array([0], dtype=cp.float64)
+    myz = cp.array(myz, dtype=cp.float64)
+
+    x = cp.array(x, dtype=cp.float64)
+    y = cp.array(y, dtype=cp.float64)
```
sbailey, line 307: Let's use `cp.asarray` to avoid copying the arrays if `x`, `y`, `edges`, and/or `myz` are already cupy arrays.

py/redrock/rebin.py
```diff
@@ -125,24 +239,150 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+
+    if (len(y.shape) == 2):
+        nbasis = y.shape[0]
+    else:
+        nbasis = 1
+
+    #Divide edges output wavelength array by (1+myz) to get 2d array
+    #of input wavelengths at each boundary in edges and
+    #use serachsorted to find index for each boundary
+    e2d = cp.array(edges/(1+myz[:,None]))
```
sbailey, line 321: given that `edges` and `myz` are already cupy arrays, I don't think this needs to be wrapped in `cp.array(...)` which would cause another memory copy, though I think it would be harmless to use `cp.asarray` here just in case a future change meant `edges` and `myz` might not be cupy arrays here.

sbailey (changes requested):
This continues the series of partial reviews with more comments/requests about the GPU implementation of `trapz_rebin`.  I haven't gotten to the meatier `zscan` yet, but I am finding it useful to establish patterns for how we want GPU/CPU compatibility and maintainability to work on these simpler examples of `lyman_transmission` and `trapz_rebin`.

py/redrock/rebin.py
```diff
@@ -98,14 +181,40 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+        use_gpu (boolean): whether or not to use GPU batch mode 
 
     Returns:
         array: integrated results with len(results) = len(edges)-1
+            In GPU batch mode, returns cp.array with shape (nz, nbin, nbasis)
+            where nbin = len(results) = len(edges)-1
+            if nz or nbasis is 1, this dimension will be squeezed,
+            e.g. for nbasis = 1 and nz = 100, the shape will be (100, nbin)
+            while for nbasis = 1 and nz = 1 the shape will be (nbin), the
+            same as CPU mode.
```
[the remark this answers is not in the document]
craigwarner-ufastro: This is done in the new version with the exception of the order of dimensions, which remains (nz, nbins, nbasis) as discussed because that is what is expected by zscan.

py/redrock/rebin.py
```diff
@@ -116,6 +225,11 @@ def trapz_rebin(x, y, xnew=None, edges=None):
     else:
         edges = np.asarray(edges)
 
+    if (myz is not None and not use_gpu):
+        raise ValueError('redshifts only allowed in GPU mode')
```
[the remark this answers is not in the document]
craigwarner-ufastro: Done in new version - with _trapz_rebin_bach (instead of _trapz_rebin_2d)

py/redrock/rebin.py
```diff
@@ -125,24 +239,150 @@ def trapz_rebin(x, y, xnew=None, edges=None):
 
     return result
 
+def trapz_rebin_batch_gpu(x, y, xnew=None, edges=None, myz=None):
```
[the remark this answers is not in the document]
craigwarner-ufastro: Done

py/redrock/rebin.py
```diff
@@ -125,24 +239,150 @@ def trapz_rebin(x, y, xnew=None, edges=None):
...
+    elif (type(myz) == np.ndarray):
+        #If myz is numpy array, process all redshifts and all bases (one at
+        #a time) and collect results in dict of 3-d numpy arrays
+        nz = len(myz)
+        result = { hs:np.empty((nz, len(wave), nbasis)) for hs, wave in dwave.items() }
+        for i, z in enumerate(myz):
+            x = (1.+z)*template.wave
+            for hs, wave in dwave.items():
+                for b in range(nbasis):
+                    result[hs][i,:,b] = trapz_rebin(x, template.flux[b], wave)
```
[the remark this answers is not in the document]
craigwarner-ufastro: Done

py/redrock/rebin.py
craigwarner-ufastro: Done

py/redrock/rebin.py
craigwarner-ufastro: Done - cp.array() has been updated to cp.asarray() in all cases in rebin.py and utils.py

py/redrock/rebin.py
craigwarner-ufastro: Done

sbailey:
@craigwarner-ufastro I tested this comparing the outputs of main with this branch and was surprised to find that ~10% of the targets changed their SPECTYPE and had considerably different redshifts, when I thought the outputs were supposed to be algorithmically identical.

Please investigate what causes these output discrepancies.  Some of the differences might be due to changes that are causing the merge conflicts (maybe from PR #218?), but I thought those were corner cases in how ZWARN was set, not something that would impact 10% of the targets.

## Test details

```
cd /global/cfs/cdirs/desi/users/sjbailey/dev/redrock/gpu
export INDIR=/global/cfs/cdirs/desi/spectro/redux/fuji/tiles/cumulative/100/20210505

#- main branch (53s + overhead)
time srun -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi -i $INDIR/coadd-0-100-thru20210505.fits -o redrock-main.fits

#- gpu branch (23s + overhead)
time srun -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi --gpu --max-gpuprocs 4 -i $INDIR/coadd-0-100-thru20210505.fits -o redrock-gpu.fits

#- gpu branch without using GPUs (40s + overhead)
time srun -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi -i $INDIR/coadd-0-100-thru20210505.fits -o redrock-nogpu.fits
```

## Example comparisons
```
from astropy.table import Table
import numpy as np

z0 = Table.read("redrock-main.fits", "REDSHIFTS")
z1 = Table.read("redrock-gpu.fits", "REDSHIFTS")

# 53 targets differ significantly
np.sum(z0['SPECTYPE'] != z1['SPECTYPE'])
np.sum(np.abs(z0['Z'] - z1['Z']) > 0.1)
```

py/redrock/utils.py
sbailey: Minor type surprise: `transmission_Lyman(zObs, lObs, use_gpu=True)` returns a cupy array if `zObs` is an array (expected), but returns a numpy array if `zObs` is a scalar.  I was expecting cupy vs. numpy array to depend only upon `use_gpu`, not upon the dimensionality of `zObs`.  I'm not sure it matters, but if it is a simple fix to ensure that use_gpu=True always returns cupy arrays, that seems more consistent.

Is this supposed to work?
```
transmission_Lyman(cupy.asarray(zObj), lObs)
```
Passing in zObj as a cupy array without explicitly setting use_gpu=True raises an exeception
```
File /global/common/software/desi/users/sjbailey/redrock/py/redrock/utils.py:262, in transmission_Lyman(zObj, lObs, use_gpu)
    260         return None
    261     lObs = tile(lObs, (zObj.size, 1))
--> 262     lRF = lObs/(1.+asarray(zObj)[:,None])
    263 T = np.ones_like(lRF)
    264 for l in list(Lyman_series.keys()):

File cupy/_core/core.pyx:1381, in cupy._core.core.ndarray.__array__()

TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.
```

This might not matter, but in the spirit of making it as transparent as possible whether we are passing around numpy vs. cupy arrays, let's fix this if it is trivial to do so (and if it isn't easy, it's ok to wait and see if this becomes a real problem or not).

py/redrock/zfind.py
```diff
@@ -33,8 +33,23 @@
...
+    Replacement for astropy.Table.sort
+    """
+    if (not colname in d):
+        raise KeyError('Key '+str(colname)+' is not in dictionary')
+    if (type(d[colname]) is not np.ndarray):
+        raise ValueError('Column '+str(colname)+' is not an np.array')
+    idx = d[colname].argsort(0).flatten()
+    for k in d.keys():
+        if (type(d[k]) is not np.ndarray):
+            raise ValueError('Column '+str(k)+' is not an np.array')
```
sbailey, line 47: Minor, pedantic: let's do the type checking and raise any exceptions before we starting modifying `d`.  Otherwise a user who catches an exception could find their input array corrupted with half the columns reordered and half not.  It might be even better to cache the reordered columns into a different dict and only update the input in-place after we are completely sure that every column is sortable (though this would use more memory).

sbailey:
Apologies for submitting an in-progress review with minor comments that step on top of the more important conversation about consistency.

@craigwarner-ufastro identified that the targets with different SPECTYPE are all GALAXY Z=0.9327... in main and STAR z=-000199... in this branch.  These are all cases of targets where the inputs have completely ivar==0 in which case redrock basically has an undefined answer.  That's bad, but not the fault of this branch.  Picking up from the previous comparison:
```
import desispec.io
coadd = desispec.io.read_spectra("/global/cfs/cdirs/desi/spectro/redux/fuji/tiles/cumulative/100/20210505/coadd-0-100-thru20210505.fits")
bad = z0["SPECTYPE"] != z1["SPECTYPE"]
for camera in ['b', 'r', 'z']:
    print(np.all(coadd.ivar[camera][bad] == 0.0))  #- True
```

Focusing on the targets that are not completely masked at input, Z agrees to O(1e-14), ZERR agrees to O(1e-12), CHI2 to O(1e-5), i.e. ok.

However, there are still ~20 targets with large DELTACHI2 discrepancies.  I suspect those are due to the differences causing the merge conflicts (from PR #218?).  @craigwarner-ufastro please resolve the merge conflicts and then let's re-compare.  I suspect this will be fine once that is resolved.

craigwarner-ufastro:
@sbailey I've confirmed the 22 additional targets with large DELTACHI2 discrepancies are in fact due to the differences causing the merge conflict in PR#218.

I've updated the `calc_deltachi2` method to match the version in that PR and that fixed the issue; however there are other discrepancies that I need to also resolve before its resolved - I suspect these changes will affect the targets that are completely masked and explain the differences in those 53 targets.  Will continue working on this until the branches are compatible and then I will push the changes out to my ongoing PR.

sbailey:
@craigwarner-ufastro thanks.  After the merge conflicts and discrepancies are resolved, one more request to finalize this PR:

Please make two more branches off of this branch, and trim them down to just the v2 algorithm and just the v3 algorithm with anything else that doesn't apply to those options removed.  I'm having a hard time tracking the final result of v2 vs. v3 and options A, B, C when some pieces are used in common and others unique to one or the other, and other pieces are completely deprecated.  Having a "just v2" branch and a "just v3" branch will be very helpful.

py/redrock/zfind.py
craigwarner-ufastro: This is resolved in the version I just pushed out - I check all columns before re-ordering any.

craigwarner-ufastro:
I have pushed out two branches: **gpu_v2only** and **gpu_v3only** that separate out only the methods necessary for those versions of my zscan algorithms.

I verified that the results (for the targets that are not all masked) agree to within floating point rounding errors.

A few notes:
1. I did end up keeping `_zchi2_one` and  `calc_zchi2_one` in both versions because although I had removed the use of `calc_zchi2_one` in the fine redshift scan, it is still used in `archetypes.py` and I don't want to break any outside code that uses it as well.

2. I got rid of all "extra" methods that were faster but we had decided against.  I did leave comments in place in the v2 and v3 algorithms describing these options, particularly for calculating the `M` and `y` arrays for now.

3. I renamed the methods e.g. `calc_zchi2_v3` -> `calc_zchi2` and `calc_zchi2_batch_v3` -> `calc_zchi2_batch`

4. In yet another datapoint in favor of V3, which is the one I have been using myself for all testing since its initial implementation, I realized that if we want to use V2, I must either create a wrapper method for `calc_zchi2_batch_v2` or refactor it to be able to be used in `fitz.py` for the fine redshift scan.  The reason is that in the way the V3 algorithm works, the `calc_zchi2_batch` method takes `spectra` and `tdata` as arguments and includes performing the dot product with the sparse matrix.  But in the V2 algorithm, that dot product is done in `calc_zchi2` and the `calc_zchi2_batch_v2` algorithm takes the output result `Tbs` as an argument.  If we choose V2, I think the way to go is to refactor this but for now I added the wrapper function `calc_zchi2_batch` which propagates to `calc_zchi2_batch_v2` as a quick and dirty method.

5. Regardless of the version chosen, logic to remove target rebalancing has been commented out so there is no additional change that needs to be done.  I had mentioned possibly removing the `ReDistTemplate` class given that its no longer needed and in fact now simply calls its superclass constructor, but for now I had simply commented out the actual rebalancing of targets starting at line 275 in `zfind.py`.

The upshots of no longer rebalancing targets are:
-  We do not run into the pickle limit for large files so larger files can be run with no problem
-  We do not run into the "ReDistTemplate not implemented for non-MPI" issue (though there may be other issues with pycuda and multiprocessing not playing nicely)
- No matter how many MPI ranks are given in the input communicator, only nGPU ranks will be actually used.
- The fastest configuration is to run with 4 GPU, 4 CPU, but running with 64 CPU and 4 GPU is only ~5s slower not including import overhead.

Timings:
```
V       CPU     GPU     Total Time      Clock Time
2       1       1       53.7            56.5
2       4       4       16.8            20.3
2       64      4       21.9            41.5
2       64      0       45.1            65.6
3       1       1       54.0            56.7
3       4       4       16.7            20.2
3       64      4       21.3            43.4
3       64      0       39.0            59.7
```

sbailey:
Thanks!  Looking at the gpu_v2only and gpu_v3only branches, I find them to be similarly complex so I will defer to your preference of v3.  Heads up: I haven't given up on making this simpler, but we can deal with that in future work so that this can be included in Iron.

I re-verified that gpu_v3only produces the same answers as main to within roundoff error (5e-4 for Z, 1e-5 for DELTACHI2), and that all targets have the same SPECTYPE and ZWARN flags, except for targets that are completely masked (see issue #224).

I also verified that this fixes #222 for GPU running on large input files.

This branch makes progress on #223 (non-MPI + GPU) such that it no longer crashes on ReDistTemplate, but now we're getting "cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorInitializationError: initialization error" when trying to use rrdesi + GPUs without MPI.  I'll update that ticket after this is merged.

I will now merge gpu_v3only -> gpu, and then merge this epic PR.

sbailey (approved):

Issue #215, opened 2022-07-12
chi2 penalty for negative emission lines


sbailey:
Currently redrock has a chi2 penalty when the PCA fit results in negative [OII], but not for other lines.  Explore whether adding chi2 penalties against [OIII], Halpha, and Hbeta would help.  Don't penalize these if they are in a redshift range that is unconstrained by the data, and subtract the model continuum first.

RaffaellaCapasso:
I added a chi2 penalty for negative [OIII], and modified the [OII] one so that it also takes into account the continuum (not necessarily showing a negative flux, but a flux lower than the continuum). See pull request [219](https://github.com/desihub/redrock/pull/219)). 

Here are some examples in which the best-fit result presents a negative [OIII] line (red line), but the second best fit does not (blue line). 

Targets in “global/cfs/cdirs/desi/users/sjbailey/redrock/example_failures”:

![39628432232943703](https://user-images.githubusercontent.com/64536272/181540583-26585320-e329-466d-8f15-d97678bd85e4.png)
![39628448234213526](https://user-images.githubusercontent.com/64536272/181540750-ba9b5689-a41c-4bc7-b183-2b52ad0e2710.png)
![39628448234216020](https://user-images.githubusercontent.com/64536272/181540814-05cdecaf-a33c-46ec-8a44-c1f0fbe6868a.png)

Fuji targets:

![39633428106185042](https://user-images.githubusercontent.com/64536272/181542352-5c97c861-52f2-42d5-b130-85e2863e02b0.png)
![39633438721969162](https://user-images.githubusercontent.com/64536272/181542900-f09866c3-b89a-4209-8227-bed6f29375e5.png)
![39633443914515633](https://user-images.githubusercontent.com/64536272/181542950-671a7be5-0719-4858-b203-d56ab8542e04.png)
![39633436100529745](https://user-images.githubusercontent.com/64536272/181542839-75ef2615-a9ff-459f-a3f9-3fb78b2a663f.png)
![39633430798930477](https://user-images.githubusercontent.com/64536272/181542568-04df83f3-94f7-4644-8b1c-bfee3909df4a.png)
![39633430803120916](https://user-images.githubusercontent.com/64536272/181542694-20f62bc4-fb1e-4b29-ade1-1feb4b11a906.png)
![39633430803124467](https://user-images.githubusercontent.com/64536272/181542777-65811751-344b-471c-87fc-ddf3f1edda21.png)

[...]

PR #219, opened 2022-07-22
chi2 penalize negative [OIII] flux

RaffaellaCapasso:
First changes to chi2 penalty for negative [OIII] line. This is connected to issue 215.

### DIFF
--- a/py/redrock/zscan.py
+++ b/py/redrock/zscan.py
@@ -140,8 +140,6 @@ def calc_zchi2(target_ids, target_data, dtemplate, progress=None, use_gpu=False)
     zchi2 = np.zeros( (ntargets, nz) )
     zchi2penalty = np.zeros( (ntargets, nz) )
     zcoeff = np.zeros( (ntargets, nz, nbasis) )
-
-    # print('test3')
     
     # Redshifts near [OII] and [OIII]; used only for galaxy templates
     if dtemplate.template.template_type == 'GALAXY':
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for this PR.  I put a few comments inline.  Please also add some info to this PR text about how often the negative [OIII] penalty causes redrock to chose a different solution, and some examples.

Note: the negative [OII] penalty was completely ad-hoc, and was added to avoid an early failure mode tested on simulations and eBOSS data.  An equivalent negative [OIII] penalty may require a different weight or offset, but that will depend upon how rare this case is, and whether the cases are switching from bad to good, or just bad to different bad, or if it is just triggering on noise fluctuations.

RaffaellaCapasso:
Thanks for the notes! I added more details to the [issue 215](https://github.com/desihub/redrock/issues/215).

sbailey:
@RaffaellaCapasso it appears that you are using the continuum when deciding if a line is negative, but not for the actual chi2 penalty, which means that a line like this would actually upweight the fit instead of penalizing it since it is negative relative to the continuum but positive relative to 0:

image

It also looks like you haven't yet put a wavelength coverage check that you are only penalizing [OIII] when it is covered by the data.

I see in issue #215 that you said "over 86 original fuji targets with negative OIII line, only 13 retained this feature in their best-fit template".  Please run tests with a much larger number of targets, including ones that did not have a negative [OIII] in the first place to confirm what fraction of targets are impacted.  e.g. "testing on 50000 targets (10 tiles), X% changed their best fit redshift..."

Please also post a plot showing your choice of [OIII] integration range and continuum ranges compared to example bad fits in restframe wavelengths, e.g. from #215 but converted to restframe and showing the integration and continuum side bands:
image

RaffaellaCapasso:
@sbailey Thanks for your comments! I have now modified the penalty so that it takes the continuum into account. I have also added a wavelength coverage check. 

Here is the plot you asked for:

Screenshot 2022-08-18 at 15 38 12

The red area is the selected [OIII] range, the green ones are the areas in which the continuum is calculated. 

I run the new code on 10 tiles, as suggested. These are the percentages of targets that changed their best-fit redshift:

- 5% of all targets (SPECTYPE == GALAXY, OBJTYPE == TGT )
- 3% of good targets (SPECTYPE == GALAXY, OBJTYPE == TGT, ZWARN==0)
- 2% of extra good targets (SPECTYPE == GALAXY, OBJTYPE == TGT, ZWARN==0 and DELTACHI2>25)

The plots below show some examples of targets that changed their best-fit redshift. Red is the original best-fit, blue is the original second best, green is the new best. In a few cases the new best-fit corresponds to the old second best (thus the blue does not appear as it's overwritten by the green one).

sbailey:
This is not converging in time for Iron, so I've removed this from the Iron project task list.

We need a more comprehensive study of why so many targets are changing; we expect ZWARN=0 DELTACHI2>25 targets to have >98% purity, so changes of 2% is large compared to the ~2% that we expect to be bad, especially since the spot checks aren't an obvious "even bad case is now good".

sbailey:
I'm closing this PR since it will be superseded by other work (significance based-penalty; see Daniel Gruen DESI data presentation on the wiki at DataSystems/Telecons/2022-11-08).

PR #220, opened 2022-10-27
keyerror fixed while reading fits file

abhi0395:
GALAXY doesn't have any subtype, so I fixed that while reading its subtype.


abhi0395 (commented):
@sbailey I think this should now work. I changed a few lines to fix the bug. The last two commits fix the bug. Please see if the changes pass the unit test.

--Abhijeet

abhi0395:
I am closing this pull request as it has become very wordy with too many commits. I'll start a new one!

Issue #225, opened 2023-01-12
re-fix treatment of bad fits when ranking minima


sbailey:
PR #211 (new GPU code) had some merge conflicts with the changes from PR #218 (avoid bad fits when ranking minima).  It appears that this wasn't completely resolved correctly, but we didn't catch it before merging #211 and the bad fits have reappeared.

A file with two example targets from sv3 dark healpix 10152 is in /global/cfs/cdirs/desi/users/sjbailey/dev/redrock/gpu/zrank/coadd-badminfit.fits.  I ran with the previous tag (0.16.0) and the latest main:
```
module swap redrock/0.16.0
time srun -n 4 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi --gpu --max-gpuprocs 4 -i coadd-badminfit.fits -o redrock-old.fits -d rrdetails-old.h5

module swap redrock/main
time srun -n 4 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi --gpu --max-gpuprocs 4 -i coadd-badminfit.fits -o redrock-gpu.fits -d rrdetails-gpu.h5
```

Comparing the top 9 fits from the rrdetails files for these two runs:
```
import h5py
import numpy as np
from astropy.table import Table

h1 = h5py.File('rrdetails-old.h5')
h2 = h5py.File('rrdetails-gpu.h5')

targetid = 39633308400748798
zfit1 = Table(np.array(h1["zfit"][str(targetid)]["zfit"]))
zfit2 = Table(np.array(h2["zfit"][str(targetid)]["zfit"]))
print(zfit1["znum", "z", "spectype", "chi2", "zwarn"])
print(zfit2["znum", "z", "spectype", "chi2", "zwarn"])
```

With tag 0.16.0 (previous main):
```
znum            z            spectype        chi2        zwarn
---- ----------------------- -------- ------------------ -----
   0      0.7465644325737828   GALAXY  8031.702453918755     4
   1     0.02090719822607346   GALAXY  8035.408617666035     4
   2   0.0012759003749705934     STAR  8116.358099282074     4
   3      1.0706405066878273      QSO  8125.152400165796     4
   4  -0.0005147214510569681     STAR  8129.311497150431     4
   5      1.0334574452161545      QSO 8133.4641371942125     4
   6       2.086404894124313      QSO  8134.274736442487     4
[...]

PR #226, opened 2023-01-13
Bugfix -  re-fix treatment of bad fits when ranking minima #225

craigwarner-ufastro:
Simple bug - np.lexsort((a,b)) results in sorting by column b first then column a.  This is counterintuitive and opposite to behavior of astropy.Table.sort((a,b)).  This has been brought in line with expected sort order by adding a boolean variable to sort_dict_by_cols, which will sort by the first column first - this is set to True by default.

### DIFF
--- a/py/redrock/test/test_zfind.py
+++ b/py/redrock/test/test_zfind.py
@@ -4,7 +4,7 @@ from io import BytesIO
 import numpy as np
 from astropy.table import Table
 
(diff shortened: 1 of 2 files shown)


sbailey:
I confirmed with re-checking example in issue #225; the sort orders now agree with original 0.16.0 tag.  Thanks!

Issue #227, opened 2023-01-17
redrock GPU test should check for GPU first


sbailey:
Redrock tests should first check if a GPU is available, and skip any tests that require a GPU.  `test_rebin` and `test_utils` tests are failing on cori for this reason.  It appears that the tests are checking if cupy is installed, but not sufficiently testing if a GPU is available (that check is inside the class setUp, but perhaps not propagating to the `cp_available` global variable).  If an entire test is to be skipped, it would be better to use @unittest.skipIf(...) wrappers so that is is reported as being purposefully skipped instead of passing.
```
======================================================================
ERROR: test_gpu_trapzrebin_uneven (redrock.test.test_rebin.TestRebin)
test rebinning unevenly spaced x for GPU vs CPU
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/redrock/main/py/redrock/test/test_rebin.py", line 122, in test_gpu_trapzrebin_uneven
    g = rebin.trapz_rebin(x, y, edges=edges, use_gpu=True)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/redrock/main/py/redrock/rebin.py", line 321, in trapz_rebin
    return _trapz_rebin_batch_gpu(x, y, edges=edges, myz=myz, result_shape=result_shape)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/redrock/main/py/redrock/rebin.py", line 376, in _trapz_rebin_batch_gpu
    edges = cp.asarray(edges, dtype=cp.float64)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/conda/lib/python3.9/site-packages/cupy/_creation/from_data.py", line 76, in asarray
    return _core.array(a, dtype, False, order)
  File "cupy/_core/core.pyx", line 2249, in cupy._core.core.array
  File "cupy/_core/core.pyx", line 2270, in cupy._core.core.array
  File "cupy/_core/core.pyx", line 2402, in cupy._core.core._array_default
  File "cupy/_core/core.pyx", line 171, in cupy._core.core.ndarray.__init__
  File "cupy/cuda/memory.pyx", line 698, in cupy.cuda.memory.alloc
  File "cupy/cuda/memory.pyx", line 1375, in cupy.cuda.memory.MemoryPool.malloc
[...]

PR #228, opened 2023-01-23
Modified unit tests to ensure that both cupy install and CUDA device

craigwarner-ufastro:
available for cp_available global var to be True.  Also modified to use unittest.skipUnless to skip GPU tests if not available.

This closes issue #227.


sbailey:
Looks good!  I empirically confirmed that the GPU tests are skipped on cori (instead of failing) and run on perlmutter (and succeed).

PR #230, opened 2023-01-26
Fix to force mpprocs = 1 when using GPU mode with multiprocessing.

craigwarner-ufastro:
GPU and multiprocessing do not play well together and mp start method must be set to spawn or forkserver to run with cupy code but the overhead inherent in these methods makes this a nonstarter. However GPU code is so fast that with --mp 1 and GPU mode is more than twice as fast as --mp 64 with no GPU.

This makes --mp 1 a special case where we forego using mp.Queue() object and follow the MPI path instead.  This also results in about a 30% speed-up on CPU with --mp 1.

Options available are:
1) MPI + CPU multiple processes
2) MPI + GPU multiple processes
3) GPU single multiprocessing process
4) CPU multiple multiprocessing processes

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -731,6 +731,12 @@ def rrdesi(options=None, comm=None):
             gpu_proc_flags = comm.allgather(use_gpu)
         else:
             gpu_proc_flags = [use_gpu, ]
+            if (mpprocs > 1):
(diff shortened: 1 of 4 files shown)


sbailey:
PR #232 was a superset of this PR.  When that was merged, it triggered github to close this redundant PR as well, but I'll also delete this branch.

PR #232, opened 2023-02-08
Remove multiprocessing from rebin

craigwarner-ufastro:
Removed the multiprocessing branch for rebinning templates because using mp.Queue resulted in slower runtime than performing `rebin_template` on a single proc.

PR #233, opened 2023-02-09
Bugfix for crashes/hangs when MPI ranks > ntargets.

craigwarner-ufastro:
rebin_template and transmission_Lyman are both passed z arrays created from master redshift list using np.array_split and those arrays can be empty if MPI ranks > ntargets.  In these cases, return output array of size 0 in nz dimension and correct other dimensions (empty array). Also one line change needed in in zscan for non-GPU case of having 0 targets for a given MPI rank to reach the t.cycle() call (in GPU mode, cycle is a no-op)


sbailey:
I tested this in combination with #232; merging both now.

PR #236, opened 2023-03-09
Fix to issue 235.  If max_gpuprocs is specified, use that value even …

craigwarner-ufastro:
…if it

results in a crash.  If not, attempt to find the number of GPUs and set max_gpuprocs equal to that instead of the number of ranks.  When using MPI, cupy.cuda.runtime.getDeviceCount() will not work if gpu-bind is used so we first attempt to check /proc and if that fails we gather the device PCI bus ids to count devices and issue a warning because this is slow (~2s for 64 ranks).  Looking at /proc should work on all Linux systems.

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -25,7 +25,7 @@ from desispec.coaddition import coadd_fibermap
 from desispec.specscore import compute_coadd_tsnr_scores
 from desispec.maskbits import fibermask
 
-from ..utils import elapsed, get_mp, distribute_work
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks.  I think this wins the award for the longest DESI branch name yet.  Thanks for all the off-list discussion and checks about what is both robust and performance optimal.  I checked that the final branch works on both GPU and non-GPU machines with default params.  Looks good.  Merging.

craigwarner-ufastro:
Haha!  I thought the same - I clicked the button on this page to fork a new branch for this issue and it created that super long name for it :)

PR #239, opened 2023-03-23
Added minwave and maxwave variables to templates.py that store wave[0]

craigwarner-ufastro:
and wave[-1] respectively.  This is used in trapz_rebin to check bin edges against input wavelength range.  Change was made after we realized that after copying template.wave and template.flux to cupy arrays, accessing them to compare to scalar on the CPU was responsible for 40% of rebin time in fine redshift scan so this eliminates need to copy to and from GPU to CPU unnecessarily.

Modified trapz_rebin to accept these as optional arguments.

Modified transmission_Lyman for a slight speed gain to ignore lines that don't overlap with wavelength range.

These changes result in a speed gain of "Finding best redshift" from 16.8s to 12.5s with 64 CPU and 4 GPU on Perlmutter with environment 23.1 and templates 0.8.

### DIFF
--- a/py/redrock/rebin.py
+++ b/py/redrock/rebin.py
@@ -17,7 +17,7 @@ block_size = 512 #Default block size, should work on all modern nVidia GPUs
 # cuda_source contains raw CUDA kernels to be loaded as CUPY module
 cuda_source = r'''
         extern "C" {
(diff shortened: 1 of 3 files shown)


sbailey:
This looks good and I confirm the timing speed up numbers, thanks!

Issue #242, opened 2023-03-27
non-reproducibility of GPU redrock


sbailey:
When run in GPU mode, redrock does not produce identical output even when run back-to-back on the same node using the same inputs.  The differences are small enough that they don't matter for final science analyses, but they are a hassle when evaluating PRs etc because you can't simply use fitsdiff to confirm that the outputs didn't change.

Example redrock runs on GPU and CPU:
```
cd $CFS/desi/spectro/redux/iron/tiles/cumulative/100/20210505

srun -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi --gpu -i coadd-0-100-thru20210505.fits -o $SCRATCH/redrock-gpu-A.fits
    
srun -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi --gpu -i coadd-0-100-thru20210505.fits -o $SCRATCH/redrock-gpu-B.fits

srun -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi -i coadd-0-100-thru20210505.fits -o $SCRATCH/redrock-cpu-C.fits
    
srun -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores \
    rrdesi_mpi -i coadd-0-100-thru20210505.fits -o $SCRATCH/redrock-cpu-D.fits
```

Differences evaluated with
```python
from astropy.table import Table
import numpy as np
from os.path import expandvars

#- GPU
a = Table.read(expandvars('$SCRATCH/redrock-gpu-A.fits'), 'REDSHIFTS')
b = Table.read(expandvars('$SCRATCH/redrock-gpu-B.fits'), 'REDSHIFTS')

#- CPU
c = Table.read(expandvars('$SCRATCH/redrock-cpu-C.fits'), 'REDSHIFTS')
d = Table.read(expandvars('$SCRATCH/redrock-cpu-D.fits'), 'REDSHIFTS')

print('GPU re-run consistency')
for col in ('Z', 'ZERR', 'CHI2'):
    ndiff = np.sum(a[col] != b[col])
    std = np.std(a[col]-b[col])
    print(f'{col=}  {ndiff=}  {std=}')

print('CPU re-run consistency')
for col in ('Z', 'ZERR', 'CHI2'):
[...]

PR #245, opened 2023-03-29
Refactored code to minimize GPU to CPU copying, resulting in big spee…

craigwarner-ufastro:
…d-ups.

The refactoring results in having two copies of several arrays (flux, wavelength, etc) - one on the GPU and one on the CPU.  It also results in several scalars such as min/max wavelength being stored on the CPU to avoid very costly single value being copied to/from GPU.  This results in a speed up on "Finding best redshift" from 12.5s to 9.3s on 64 CPU / 4 GPU and from 10.0s to 7.2s on 4 CPU / 4 GPU, bringing it closer in line with the 6.7s for 64 CPU.

### DIFF
--- a/py/redrock/zscan.py
+++ b/py/redrock/zscan.py
@@ -488,6 +488,7 @@ def calc_batch_dot_product_3d3d_gpu(a, b, transpose_a=False, fullprecision=True)
         ncols = cp.int32(a.shape[1])
     transpose_a = cp.int32(transpose_a)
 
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks.  I independently confirmed that
1. this branch is faster than main
2. repeated runs of this branch are exactly reproducible (unlike main), and they agree with main to the precision that main agrees with itself under multiple runs (fixes #242)

Issue #246, opened 2023-04-05
Unit test failure in test_gpu_zscan


marcelo-alvarez:
On Perlmutter unit test test_gpu_zscan fails:
```
% source /global/common/software/desi/desi_environment.sh main
% pytest py/redrock/test/test_zscan.py
...
========================================== FAILURES ==========================================
__________________________________ TestZScan.test_gpu_zscan __________________________________

self = <redrock.test.test_zscan.TestZScan testMethod=test_gpu_zscan>

    @unittest.skipUnless(cp_available, "Skipping test_gpu_zscan because no GPU found.")
    def test_gpu_zscan(self):
        z1 = 0.2
        z2 = 0.25
        seed = np.random.randint(2**31)
        print('GPU TEST: Using random seed {}'.format(seed))
        np.random.seed(seed)

        t1 = util.get_target(z1); t1.id = 111
        t2 = util.get_target(z2); t2.id = 222
        dtarg = DistTargetsCopy([t1, t2])

        # Get the dictionary of wavelength grids
        dwave = dtarg.wavegrids()

        # Construct the distributed template.
        template = util.get_template(redshifts=np.linspace(0.15, 0.3, 50))
        dtemp = DistTemplate(template, dwave)

        results_a = calc_zchi2_targets(dtarg, [ dtemp ], use_gpu=False)
>       results_b = calc_zchi2_targets(dtarg, [ dtemp ], use_gpu=True)

py/redrock/test/test_zscan.py:162:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
py/redrock/zscan.py:851: in calc_zchi2_targets
    calc_zchi2(targets.local_target_ids(), targets.local(), t, use_gpu=use_gpu)
py/redrock/zscan.py:722: in calc_zchi2
    (zchi2[j,:], zcoeff[j,:,:]) = calc_zchi2_batch(target_data[j].spectra, tdata, weights, flux, wflux, nz, nbasis, dtemplate.template.solve_matrices_algorithm, use_gpu)
py/redrock/zscan.py:537: in calc_zchi2_batch
    Tbs = batch_dot_product_sparse(spectra, tdata, nz, use_gpu)
py/redrock/zscan.py:258: in batch_dot_product_sparse
    return _batch_dot_product_sparse_gpu(spectra, tdata)
[...]

PR #247, opened 2023-04-05
Fixed unit test test_gpu_zscan to create GPU version of template when

craigwarner-ufastro:
running on GPU so that arrays are correctly CUPY.


marcelo-alvarez:
@craigwarner-ufastro thanks. The way that `calc_zchi2_targets` is designed and used by the code requires that its `templates` argument be in GPU mode when `use_gpu=True`, so your changes here fix #246. I have verified this is the case with my own branch. I'll merge now.

PR #253, opened 2023-09-19
Abhijeet per camera gpu

craigwarner-ufastro:
Use batch rebin, transmission_Lyman, and calc_zchi2 operations.

Speed gains: without archetypes, redrock on 4 GPU / 4 CPU takes 14.8s
reported total run time, 7.3s of which is in the fine redshift scan.
Comparatively with 64 CPU and 0 GPU the base redrock runs in 40.0s
reported total run time with 6.7s spent in fine redshift scan.

Adding the base archetypes option (without per-camera or nearest
neighbor) raises CPU times to 63.8s overall and 28.1s in fine z scan so
about 60% increase overall and about 4x slower in fine z scan.

With the new code the "batch" CPU mode slightly improves this to 60.0s
and 24.2s.

With the new GPU code, it runs on 4 GPU / 4 CPU in 22.8s total and 14.3s
for fine z scan, a 50% overall increase but only 2x speed increase in
the fine z scan, a big improvement from CPU times.

Also updated transmission_Lyman to return None when given scalar z to
match behavior when given an array of redshifts, in the case where the
wavelength range is not affected by Lyman transmission.  There is no
need in this case to calculate an array of all ones and then
additionally multiply the rebinned data by it.

Have yet to update --per-camera and -n_nearest options so right now
there are placeholders so that it does not crash that simply loop over
the existing CPU-mode logic.  These will be updated shortly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,7 @@ Or::
 
     git clone https://github.com/desihub/redrock-archetypes.git
 
-In summary, the archetypes method uses a combination of physical galaxy spectra and Legendre polynomials to construct a new set of templates and then solve for the coefficients using the bounded value least square method. The method solves for the coefficients of the Legendre polynomials in each camera (b, r, z cameras of desi, ``--per-camera`` keyword is introduced for that).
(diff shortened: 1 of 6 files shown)


craigwarner-ufastro:
Timing update - rows after the blank line are new results
```
Mode            CPU     GPU     Findz   Total   Clock
Orig             4      4        7.3    14.8    18.6
Orig            64      0        6.7    40.0    49.1
Arch only       64      0       28.1    63.8    71.7
Arch only batch 64      0       24.2    60.0    68.0
Arch only GPU    4      4       14.3    22.8    27.4
Arch+PC+NN      64      0       40.1    76.4    93.5
Arch+PerCam      4      4       376.7  389.0   400.6

Arch only batch 64      0       24.3    60.1    67.9
Arch only GPU    4      4       12.7    21.8    26.0
Arch+NN batch   64      0       24.6    60.3    69.7
Arch+NN GPU      4      4       15.5    23.9    28.1
```

craigwarner-ufastro:
Added NNLS to `solve_matrices` and restored `per_camera_coeff_with_least_square_batch` in zscan.py from my previous version.

sbailey:
changes in this PR were incorporated into PR #255, so closing this one without merging.

PR #255, opened 2023-11-01
Abhijeet per camera prior gpu

abhi0395:
This branch combines the PCA with archetypes and a per-camera modelling approach. It is an important step in improving the current Redrock, particularly for galaxy spectral fitting and estimating their redshifts. Thanks to @craigwarner-ufastro for importing some of the time-consuming steps to GPUs; it has made a significant difference.
@craigwarner-ufastro I merged your branch locally and added prior option to it so this branch includes all your GPU updates to the archetype method. So please feel free to close your PR.

In the current form, redrock occasionally yields nonphysical model fits for galaxies and struggles to account for errors introduced by the spectral reduction pipeline adequately. To address this persistent challenge, our proposed approach combines redrock with galaxy archetypes and introduces per-camera modelling for spectral fitting and redshift estimation.

I have run extensive tests on many tiles, including SV3 visual tiles, repeat observations and large test runs on iron data and compared them with results from **iron data reduction redrock**. The method has shown several important improvements over redrock, such as mitigating non-physical/negative emission line fitting. It also reduces the number of sky fibers having very robust redshifts by a significant fraction (25-50%) and reduces the spurious clustering of redshifts in the redshift vs fiber plane without changing any quality cuts that redrock currently uses. It also shows a slight increase in the redshift success rate for most target classes.

There are a few new options added to `rrdesi`:

- `--archetypes`: file/directory containing the archetypes
- `--per_camera`: to do the spectral fitting in each camera (i.e. b, r, z)
- `-deg_legendre`: number of Legendre polynomials to be used in archetype mode
- `--nminima`: number of redshifts on which archetype model should be implemented to estimate the final redshifts
-  `--prior_sigma`: variance to be added in the final linear equation before solving for the coefficients

The new archetypes are saved in `/global/cfs/cdirs/desi/users/abhijeet/new-archetypes/`

The most important bash commands to run `rrdesi` with `archetypes` are detailed in the README file so anyone can run and analyze the outputs. For the current test runs, we have used the following bash commands:

**Without priors**:
`rrdesi -i <spectra_file> --archetypes <archetype_dir or archetype_file> -o <output_file> -d <details_file.h5> -deg_legendre 2 --nminima 9 --per-camera`

**With priors**:

`rrdesi -i <spectra_file> --archetypes <archetype_dir or archetype_file> -o <output_file> -d <details_file.h5> -deg_legendre 2 --nminima 9 --per-camera --prior_sigma 0.1`

The main directory (`$ABHIJEET_TEST_DIR`) where test results are saved: 
`$ABHIJEET_TEST_DIR=/global/cfs/cdirs/desi/users/abhijeet/large_test_run `

- Results **without prior** are in : `REF_DIR = ${ABHIJEET_TEST_DIR}/no_prior_runs`

- Results **with prior** are in : `REF_DIR = ${ABHIJEET_TEST_DIR}/prior_runs`

The subdirectory structures are the following (in both directories):

- **visual-tiles**: `${REF_DIR}/visual_tiles`
- **pernight tiles (repeat observations)**: `${REF_DIR}/pernight`
- **iron tiles** (large runs from iron data nights): `${REF_DIR}/iron_nights`

For comparison, the **iron data reduction** with redrock is at: `/global/cfs/cdirs/desi/spectro/redux/iron/tiles/cumulative`

Many thanks to @julienguy, @sbailey, John and members of the DESI data team for their help, suggestions and feedback on the progress.


py/redrock/external/desi.py
```diff
@@ -580,6 +592,9 @@ def rrdesi(options=None, comm=None):
 
     parser.add_argument("--nminima", type=int, default=3,
         required=False, help="the number of redshift minima to search")
+    
+    parser.add_argument("--prior_sigma", type=float, default=None,
+        required=False, help="sigma to add as prior in solving linear equation, 1/sig**2 will be added")
```
sbailey, line 597: The command line options are getting confusingly similar, e.g. `--prior_sigma` has nothing to do with `--priors`.  I suggest that we standardize all archetype-related options to start with `--archetype`, e.g.
```
-deg_legendre -> --archetype-legendre-degree
--per-camera  -> --archetype-legendre-percamera
--prior_sigma -> --archetype-legendre-prior
-n_nearest    -> --archetype-nnearest (?)
```
We might also want to consider a single `--archetype-legendre` that turns on the other 3 legendre-related terms with the recommended defaults.

Let's also rename `-nz` (number of finer redshift to be used around best fit redshifts) to `--zminfit_npoints` or something similar to try to make the option name clearer that it is number of points fit around each minimum, not the total number of redshifts to fit.

py/redrock/fitz.py
```diff
@@ -107,7 +107,25 @@ def minfit(x, y):
...
+    wave = np.concatenate([ w for w in dwave.values() ])
+    wmin = wave.min()
+    wmax = wave.max()
+    legendre = { hs:np.array([scipy.special.legendre(i)( (w-wmin)/(wmax-wmin)*2.) for i in range(nleg)]) for hs, w in dwave.items() }
+
+    return legendre
+
+def prior_on_coeffs(n_nbh, deg_legendre, sigma):
+
+    nbasis = n_nbh+deg_legendre*3 # 3 desi cameras
```
sbailey, line 121: Please explore how to avoid hardcoding 3 as the number of DESI cameras, e.g. by inspecting the number of different wavelength grids represented in the input spectra.  Redrock has been used by eBOSS (with 2 cameras coadded into 1 before fitting), and by WEAVE (2 cameras), and could also be used on DESI with coadded cameras.  Let's try to preserve that generality.

py/redrock/utils.py
```diff
@@ -255,6 +255,11 @@ def transmission_Lyman(zObj,lObs, use_gpu=False):
     min_wave = 0
     if (np.isscalar(zObj)):
         #zObj is a float
+        min_wave = lObs.min()/(1+zObj)
+        if (min_wave > Lyman_series['Lya']['line']):
+            #Return None if wavelength range doesn't overlap with Lyman series
+            #No need to perform any calculations in this case
+            return None
```
sbailey, line 262: @craigwarner-ufastro did this optimization come from you?  How significant an improvement was it?  To me it seems more convenient to always return the transmission, even when it is all ones, so that the end user doesn't have to check the return type every time.  Or perhaps this optimization could become an opt-in option, to be used internal to the code, but without imposing it on other users of transmission_Lyman.

py/redrock/zfind.py
```diff
@@ -546,13 +588,20 @@ def zfind(targets, templates, mp_procs=1, nminima=3, archetypes=None, priors=Non
             del allresults[tid]['meta']
 
         # Standardize column sizes
-        if allzfit['subtype'].dtype != '<U20':
-            allzfit.replace_column('subtype', allzfit['subtype'].astype('<U20'))
+        if allzfit['subtype'].dtype != '<U32':
+            allzfit.replace_column('subtype', allzfit['subtype'].astype('<U32'))
```
sbailey, line 592: What's the case where SUBTYPE needs to be 32 characters instead of 20?  I don't see any subtypes in the current files that are that long, and this changes the datamodel of the output files.

py/redrock/zfind.py
```diff
@@ -546,13 +588,20 @@ def zfind(targets, templates, mp_procs=1, nminima=3, archetypes=None, priors=Non
...
 
         if allzfit['spectype'].dtype != '<U6':
             allzfit.replace_column('spectype',allzfit['spectype'].astype('<U6'))
 
-        maxcoeff = np.max([t.template.nbasis for t in templates])
+        if not per_camera:
+            maxcoeff = np.max([t.template.nbasis for t in templates])
+        else:
+            ncam = 3 # three cameras of DESI:: b, r, z
```
sbailey, line 600: See prior comment about avoiding hardcoding that DESI has 3 cameras.  And if we do hardcode it, we shouldn't hardcode it in 2 places.

sbailey (changes requested):
When running without archetypes, this branch runs in the same time as current main and produces the same output, except for changing the SUBTYPE size from 20 to 32 characters.  Is that change necessary or leftover from something else?

When running with archetypes without legendre+priors, the overall runtime is ~40% slower, which is unfortunate but could be acceptable (spending previous GPU speedups to enable better algorithms).

When running with archetypes and legendre+priors, the runtime becomes 12x slower (!).  That doesn't have to be a blocking factor for merging, but it is effectively a blocking factor for using those options, so we'll need to revisit that.

I put various comments inline.  Most are minor and/or apply only to archetypes, but I would like to standardize / dis-ambiguate the option names before merging.

I think we should also resolve + merge #259 to get various GPU changes into this branch before merging this branch into main.

Heads up: I also haven't reviewed the output formats yet.  I think we should more cleanly separate the coefficients from the original PCA (or NMF) scan from the archetype coefficients from the legendre coefficients, but I don't have a specific suggestion yet.  That might be a post-merge update.  Thinking.

sbailey:
Looking at /global/cfs/cdirs/desi/users/abhijeet/new-archetypes/rrarchetype-galaxy.fits, there is an unnamed HDU 2 with a binary table that has units and special characters in the column names:
```
  column info:
    LOGMSTAR [M_sol]
                        f8  
    LOGSSFR [yr^-1]     f8  
    AV_ISM [mag]        f8  
    TEMPLATEID          i4  
    SUBTYPE            S23  
```
Although that is technically valid, it is somewhat unusual, and would probably be better to have column names like "LOGMSTAR", "LOGSSFR", and "AV_ISM" and put the units in TUNITnn keywords or in the comments.  Please also add an EXTNAME to HDU 2.

py/redrock/utils.py
craigwarner-ufastro: @sbailey yes this was from me - and if I recall it was a very significant difference because most non-QSO templates don't overlap with the Lyman series.  So having to create an array of all ones and multiply thousands of times for each target caused a not insignificant and unnecessary slowdown.  We could do the opt-in option though without hurting any performance.

py/redrock/archetypes.py
```diff
@@ -63,6 +86,45 @@ def rebin_template(self,index,z,dwave,trapz=True):
         else:
             return {hs:self._archetype['INTERP'][index](wave/(1.+z)) for hs, wave in dwave.items()}
 
+    def rebin_template_batch(self,z,dwave,trapz=True,dedges=None,indx=None,use_gpu=False):
+        """
+        """
```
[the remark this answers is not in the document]
craigwarner-ufastro: @sbailey the batch method was from me - I don't think I omitted the docstring but maybe it got accidentally deleted somehow?  I'll re-add it.

abhi0395:
@sbailey 
https://github.com/desihub/redrock/pull/255#issuecomment-1830900168

Thanks for the suggestions. I have updated the file now.

py/redrock/zfind.py
abhi0395: I changed it to 'U32' if we ever use the nearest neighbours. In that case, the subtypes can be longer.

py/redrock/zfind.py
abhi0395: Yes, I agree, it is better to add ncam in rrdesi (with default value 3).

py/redrock/fitz.py
abhi0395: I think, we can think of adding ncam in rrdesi. Fixing the default value to 3.

py/redrock/utils.py
craigwarner-ufastro: Done - new argument is `always_return_array`

py/redrock/archetypes.py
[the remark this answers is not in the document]
craigwarner-ufastro: Done. @abhi0395 can check that my docstring is correct.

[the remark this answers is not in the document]
abhi0395: This looks good to me.

py/redrock/external/desi.py
abhi0395: I have cleaned up the arguments now. I have added a flag `--archetype-legendre` that will run archetypes with default values. Otherwise user can also provide desired arguments.

py/redrock/fitz.py
abhi0395: Fixed now, code can calculate number of camera using spectra information

sbailey (approved):

py/redrock/zfind.py
sbailey: This is now auto-derived from the input spectra files.

PR #260, opened 2023-12-04
Support for Fastspec PCA templates

sbailey:
I'm opening this PR with changes from John and Craig to add the Inoue14 IGM model with GPU support, plus command line options to specify redshift scan ranges per template.  However, to maintain compatibility with current templates and runs, I'm leaving the Calura12/Irsic13 IGM model as the default.  In a separate branch+PR I'll refactor how IGM models are handled to support backwards compatibility of IGM models while allowing Inoue14 for future templates.


sbailey:
The new Inoue IGM code is not covered by unit tests yet; I will add that as part of a separate IGM refactor branch+PR.  I have verified that this branch produces identical outputs as main when run with default parameters.

PR #263, opened 2024-01-02
Fix #261

### LINKED ISSUE #261: test_io should use temp directory
redrock.test.test_io.TestIO currently writes temp files into the top-level directory of the git clone.  This risks leaving temp files behind upon test crashes, and also means that they can't be run from a batch node when the clone is on /global/common/software/ which is mounted readonly.

TODO: use tempfile.mkdtemp() to get a temp directory to write test files to instead.

moustakas:
Write test file(s) in `test_io.py` to temporary directory; should address #261.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,9 @@ redrock Change Log
 0.18.1 (unreleased)
 -------------------
 
-* No changes yet.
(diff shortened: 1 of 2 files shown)


sbailey (approved):
Looks good, thanks!

PR #265, opened 2024-01-04
added default --nminima for archetypes

abhi0395:
Implemented a default value for the `--nminima` argument in the archetype method, which will be invoked when redrock runs in archetype mode. Additionally, incorporated logic to allow the user to overwrite the default value by providing the `--nminima` argument.

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -605,7 +605,7 @@ def rrdesi(options=None, comm=None):
     parser.add_argument("-n", "--ntargets", type=int,
         required=False, help="the number of targets to process in each file")
 
-    parser.add_argument("--nminima", type=int, default=3,
+    parser.add_argument("--nminima", type=int, default=None,
         required=False, help="the number of redshift minima to search")
     
     parser.add_argument("--allspec", default=False, action="store_true",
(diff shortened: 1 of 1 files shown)


py/redrock/external/desi.py
```diff
@@ -848,9 +850,15 @@ def rrdesi(options=None, comm=None):
...
             archetype_legendre_percamera = False
+            nminima = args.nminima 
         else:
             if comm_rank == 0 and args.archetypes is not None:
-                print('Will be using default archetype values.')
+                print('Will be using default archetype values.') 
+                if args.nminima==parser.get_default('nminima'):
+                    nminima = 9 #default for archetype
+                else:
+                    nminima = args.nminima
```
sbailey, line 860: This if/else is inside a "if comm_rank == 0..." block, so only MPI rank 0 will set this.  I think it would be better to handle this earlier in the code with
```python
# default=None to be able to tell if user sets it
parser.add_argument("--nminima", type=int, default=None,
        required=False, help="the number of redshift minima to search")
...
# if user sets --nminima, use that
if args.nminima is not None:
    nminima = args.nminima
# otherwise use different defaults for archetypes vs. non-archetypes
elif args.archetypes is None:
    nminima = 3    # default for non-archetypes
else:
    nminima = 9    # default for archetypes
```

abhi0395: Yes, I am doing it now, I just realized that, thanks!

PR #266, opened 2024-01-04
adjust coeff size for legendre only for archetypes

sbailey:
This PR is a followup to #255 (archetypes), fixing a bug where deg_legendre was being used to adjust the size of the COEFF array even if archetypes were not being used and thus no legendre polynomials either.

I didn't catch this while testing PR #255 because I had been using galaxy templates with ncoeff=10, which was greater than ncamera*deg_legendre+1 = 3*2+1 = 7, and so the coeffs arrays remained size 10.  However, when running the full pipeline, the QSOQN afterburner re-runs Redrock with only the quasar templates with ncoeff=4.  Redrock was incorrectly padding this upward to 7 coefficients, which confused the afterburner expecting only 4.

@abhi0395 can you double check this?  We need to get a fix in place before observations tonight.  Thanks.

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -851,9 +851,9 @@ def rrdesi(options=None, comm=None):
         dwave = targets.wavegrids()
         
         ncamera = len(list(dwave.keys())) # number of cameras for given instrument
-        if args.archetypes_no_legendre:
+        if args.archetypes_no_legendre or args.archetypes is None:
             if comm_rank == 0:
-                print('--archetypes-no-legendre argument is provided, will turn off all the Legendre related arguments')
+                print('no archetypes or --archetypes-no-legendre; will turn off all the Legendre related arguments')
             archetype_legendre_prior = None
             archetype_legendre_degree =0
             archetype_legendre_percamera = False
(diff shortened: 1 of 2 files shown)


abhi0395 (approved):
Looks good!

sbailey:
I reviewed this in person with Abhijeet, and he made the good suggestion of also turning off legendre terms in the rrdesi wrapper when archetypes are not used.  I added that.

Tested with:
```
cd $CFS/desi/spectro/redux/iron/tiles/cumulative/1000/20210517

# no archetypes, all templates, produces same answer as main from before #255 (redrock-main.fits)
time srun -n 4 -c 32 rrdesi_mpi --gpu -i coadd-0-1000-thru20210517.fits \
        -o $SCRATCH/redrock-all.fits

# running with only qso-HIZ template; output has 4 coeff (like before), not 7
time srun -n 4 -c 32 rrdesi_mpi --gpu -i coadd-0-1000-thru20210517.fits \
        -t $RR_TEMPLATE_DIR/rrtemplate-qso-HIZ.fits \
        -o $SCRATCH/redrock-qso.fits
```

For the record: in the long term I'd like to separate out the Legendre logic from the archetype logic so that we could use Legendre polynomial terms for the initial PCA/NMF scan as well.  I'd also like to separate these into separate columns for the PCA/NMF coefficients, the archetype coefficients, and the legendre coefficients instead of using a single COEFF column with its meaning depending upon the context of what options were used.

Issue #257, opened 2023-11-10
Inoue IGM transmission does not accept scalar redshifts, even though it claims to


dylanagreen:
On the branch `fastpsec-pca-templates`, the full IGM transmission curve using the model from Inoue+ 14 is implemented for future use in redrock. A later GPU-ified version iterated upon this and now exists as the most up to date version. This function is (on line 714 of redrock.utils) `transmission_IGM` with the old version in `transmission_IGM_old`. The docstring for `transmission_IGM` (and `transmission_IGM_old`) claims:
> 3-> For legacy, it is still supported to pass zObj as a float and
    in this case, the return value will be a 1-d numpy array (nlambda).

This, however, does not seem to be true. The following minimal reproducible example fails:

```python
from redrock.utils import transmission_IGM
xmin = 2500
xmax = 6500
x = np.linspace(xmin, xmax, num=50)
z = 3.5
transmission_IGM(z, x)

----------------------------
    769 i = min_wave < 1220.
--> 770 T[i, :] = IGM.full_IGM(asarray(zObj[i]), lObs[i,:], use_gpu=use_gpu)
    771 if np.isscalar(zObj) and use_gpu:
    772     T = asarray(T)

> TypeError: 'int' object is not subscriptable
```

Note that I believe there are in fact two bugs here, the second hidden by the first. The first is outlined in the error message above, where `zObj[i]` is attempting to index a scalar, and the check for scalar `zObj` is only after running the full IGM. There is a second bug, namely, even if `zObj` is passed correctly to `full_IGM` there is an entirely separate instance inside the object class that also implicitly indexes on the input redshift, even if it is a scalar:

```python
from redrock.utils import Inoue14
xmin = 2500
xmax = 6500
x = np.linspace(xmin, xmax, num=50)
z = 3.5
igm = Inoue14()
igm.full_IGM(z, x)

----------------------------
    371 else:
    372     arrexp = np.exp
--> 374 tau_LS = self.tLSLAF(z, lobs, use_gpu=use_gpu) + self.tLSDLA(z, lobs, use_gpu=use_gpu)
    375 tau_LC = self.tLCLAF(z, lobs, use_gpu=use_gpu) + self.tLCDLA(z, lobs, use_gpu=use_gpu)
    377 ### Upturn at short wavelengths, low-z
    378 #k = 1./100
    379 #l0 = 600-6/k
    380 #clip = lobs/(1+z) < 600.
[...]

PR #271, opened 2024-01-13
Versioned templates, NMF support, updated IGM models

sbailey:
This PR adds flexibility to keep multiple versions of the same SPECTYPE/SUBTYPE templates in the redrock-templates directory, while having each template file specify which intergalactic medium (IGM) model to use and whether it should be solved using PCA (cp/np.linalg.solve) or NMF (scipy.optimize.nnls) with only positive coefficients.  It is backwards compatible with current redrock-templates directories, while allowing more flexibility for the future

## New templates directory ##

An example of the new templates directory is $CFS/desi/users/sjbailey/code/redrock-templates, including
```
default_templates.txt
iron_templates.txt
test_templates.txt
rrtemplate-galaxy-None-v2.6.fits
rrtemplate-qso-HIZ-v1.0.fits
rrtemplate-qso-HIZ-v1.1.fits
rrtemplate-qso-HIZ-v2.0b1.fits
rrtemplate-qso-LOZ-v1.0.fits
rrtemplate-qso-LOZ-v2.0b1.fits
rrtemplate-qso-None-v0.1.fits
rrtemplate-star-*-*.fits
```

The file `default_templates.txt` specifies which templates to use by default, but other files like `test_templates.txt` can specify other sets to use, e.g. for testing newer templates (like `rrtemplate-qso-*-v2.0b1.fits` which are NMF-based templates)

New template file keywords:
  * RRIGM = which IGM model to use: Calura12, Kamble20, Inoue14
  * RRMETHOD = 'PCA' or 'NMF'
  * I also added the missing VERSION keywords to the star templates

This is a draft directory for testing; I would not commit all of the test templates yet.  The code in this PR can use either this new directory structure, or an existing redrock-templates like /global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/redrock-templates/main/ .

## New IGM structure ##

Lyman / IGM absorption code has been moved out of redrock.utils into a separate redrock.igm module.  This currently supports 3 models:
  * Calura12 (used by Fuji and Iron)
  * Kamble20 (same functional form, updated coeffs, used by @abrodze for updated qso-HIZ-v1.1 templates post-iron)
  * Inoue14 (planned to be used for future templates)

`redrock.igm.transmission_Lyman(redshifts, obswavelengths, model=model, ...)` is a standardized wrapper around all 3 of those models, as well as standardizing scalar vs. vector input, use_gpu or not, always_return_array or not.  Tests have been significantly expanded to check all those cases, e.g. fixing #257.

## Testing ##

**Reference run with main and current templates**
```
cd $CFS/desi/spectro/redux/iron/tiles/cumulative/1000/20210517

srun -n 4 -c 32 rrdesi_mpi --gpu -i coadd-0-1000-thru20210517.fits \
    -o /pscratch/sd/s/sjbailey/rr/redrock-main.fits
```

**New code, current templates**
```
srun -n 4 -c 32 rrdesi_mpi --gpu -i coadd-0-1000-thru20210517.fits \
    -t /global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/redrock-templates/main/ \
    -o /pscratch/sd/s/sjbailey/rr/redrock-pr-oldtemplates.fits
```

**New code, new template directory, but same underlying templates**
```
srun -n 4 -c 32 rrdesi_mpi --gpu -i coadd-0-1000-thru20210517.fits \
    -t $CFS/desi/users/sjbailey/code/redrock-templates \
    -o /pscratch/sd/s/sjbailey/rr/redrock-pr.fits
```

**New code, new NMF-based QSO templates**
```
srun -n 4 -c 32 rrdesi_mpi --gpu -i coadd-0-1000-thru20210517.fits \
    -t $CFS/desi/users/sjbailey/code/redrock-templates/test_templates.txt \
    -o /pscratch/sd/s/sjbailey/rr/redrock-prtest.fits
```
(For demo purposes, not yet better actual redshifts...)

fitsdiff confirms that redrock-pr-oldtemplates.fits and redrock-pr.fits match redrock-main.fits except for header keywords tracking input templates versions/locations.  redrock-prtest.fits differs as expected, but has strictly positive coefficients for anything with SPECTYPE=QSO due to the NMF templates.

Using previous templates and/or new templates with Calura12 or Kamble20 IGM models runs in the same amount of time as before (~36 seconds for 500 spectra).  Using NMF is a bit slower (~40 sec), but switching to the Inoue14 model is substantially slower (~55 sec).  This was known from off-list discussions when the Inoue model was first added (PR #260) and could be the focus of a future optimization+PR.

## Reviewing ##

@dylanagreen please review, especially checking whether my reorganization of the Inoue model still agrees with your implementation.  Comments welcome from @abrodze, @moustakas, @abhi0395, @craigwarner-ufastro, or others as well.

Since this is backwards compatible with current templates and results, I think this is safe to merge after someone else verifies that.  But it lays the groundwork for testing future templates.

## Future work beyond this PR ##

* finalize and commit new redrock-templates/ structure
  * Note: the new code is backwards compatible with the old redrock-templates/ directory structure, but the current main code is not future-compatible with the new redrock-templates/ structure, so we want to do this carefully.
* improve efficiency of Inoue14 implementation
* fix `redrock.templates.Template.eval` and `redrock.templates.eval_model` to also include the IGM model.  Neither of these functions is actually used by the core redrock code but they were provided for end-user convenience, but they are currently wrong (and they were wrong before this PR as well).
* augment `eval_model` (or new function) to auto-derive which templates to load based upon header metadata recording the template versions used in a previous run.  An example use case is rendering the models for Fuji or Iron fits to the data, even when we are using newer templates for current runs.
* Add support for archetypes to also specify their IGM model (current galaxy archetypes aren't designed for high-z anyway, so this doesn't matter yet).


py/redrock/igm.py
```diff
@@ -0,0 +1,447 @@
...
+    1 -> everything is transmitted (medium is transparent)
+    0 -> nothing is transmitted (medium is opaque)
+
+    This method will handle 3 options:
+    1 -> For GPU mode, zObj is an array of all z for a template and the return
+    value will be a cupy array (nz x nlambda)
+    2-> In CPU mode, it can auto-detect if zObj is a numpy array and if so,
+    again, all z will be processed as a vector and the return value
+    will be a numpy array (nz x nlambda)
+    3-> For legacy, it is still supported to pass zObj as a float and
```
dylanagreen, line 381: This line in the docstring here is not strictly true. Scalar z only works for Calura/Kamble models, for Inoue scalar z will break due to #257. Is it worth pursuing a fix to the aforementioned issue or at the very least updating the docstring before merging?

dylanagreen (approved):
I made a single comment on one possible gotcha for the end user but otherwise have encountered no issues with this version of redrock. I have tested it with two different sets of NMF templates now, for 40k spectra each with no issues, and have independently verfied that the output of transmission_Lyman(... model="Inoue14"...) matches my own independent Inoue14 implementation to 1e-16 or less, so I'm signing off on these changes (for whatever that is worth)

moustakas:
FWIW I'm a few days away from being able to test the various hooks of this branch myself, but I'm also happy to report any problems as new tickets.

py/redrock/igm.py
sbailey: Thanks for catching that.  I actually had updated the function for full support of all combinations of vector vs. scalar zObj and lObs for both CPU and GPU along with unit tests to prove that it worked (8 combinations!) ... and then forgot to update the docstring!  I've corrected the docstring.

Note that the wrapper function transmission_Lyman is the only one with full support for all combos.  The underlying functions transmission_IGM_Inoue14 and transmission_Lyman_CaluraKamble now both require array inputs and produce 2D array outputs, and the wrapper function handles up/down shaping those to scalars, 1D arrays, and 2D arrays.

Issue #281, opened 2024-02-13
problematic QSO / GALAXY fit


moustakas:
At the data systems call today I presented this oddball. Redrock ranks a QSO model as the top two fits, both at the wrong (VI) redshift, but the sharpness of the chi^2 scan at the minimum suggests that something is amiss. The GALAXY fit (shown in gray) is at the right redshift, looks like a clearly better fit, but is ranked as the third-best minimum.

![image](https://github.com/desihub/redrock/assets/1431820/8deb1148-986e-461b-92b7-7237e5a65f56)

To reproduce:
```
rrdesi -i /global/cfs/cdirs/desi/users/ioannis/fastspecfit/redrock-templates/stacks/redux-templates-NMF-0.2-zscan01/vitiles/coadd-4-80613.fits \
  --targetids 39633348334715174 -o zbest.fits -d zdetails.h5 \
  -t /global/cfs/cdirs/desi/users/ioannis/fastspecfit/redrock-templates/stacks/rrtemplates/NMF-0.2 
```
And then (third minimum is the correct one):
```
from redrock.results import read_zscan
zs, zf = read_zscan('zdetails.h5')
zf
<Table length=9>
     targetid                z                     zerr          zwarn        chi2                               zz                       ... npixels spectype subtype ncoeff  znum     deltachi2
      int64               float64                float64         int64      float64                         float64[15]                   ...  int64    str6     str3  int64  int64      float64
----------------- ----------------------- ---------------------- ----- ------------------ ----------------------------------------------- ... ------- -------- ------- ------ ----- ------------------
39633348334715174      0.6854235253876094  6.435125524024464e-06     0  93344.39492797852        0.6834076602245445 .. 0.6872883156635221 ...    7925      QSO     LOZ      4     0 12612.389946937561
39633348334715174     0.28587434510372395  6.537056198330875e-06     0 105956.78487491608       0.2843674750024283 .. 0.28732824781181976 ...    7925      QSO     LOZ      4     1   6630.57647138834
39633348334715174      0.1497488356821888   8.95398520026666e-06     0 112587.36134630442      0.14874353633682325 .. 0.15033168098113836 ...    7925   GALAXY             40     2 2468.6016489243048
39633348334715174     0.15001886380785487 1.0588134929353917e-05     0 114171.00924909115      0.14865418458569857 .. 0.15130210595034432 ...    7925      QSO     LOZ      4     3   884.953746137573
[...]

PR #282, opened 2024-02-20
add --negflux-nsig option to mask significantly negative flux

sbailey:
@moustakas this PR adds an rrdesi --negflux-nsig option (default 5) to mask significantly negative data.

There is still some ongoing discussion about the case of sky mis-subtraction positive-negative fluctuations — masking only the negative here risks leaving positive peaks behind to fit false emission lines.  So lets *not* merge this yet, but this provides a branch to test and a place for comments.

Default rrdesi/rrdesi_mpi usage is unchanged, and it will apply a mask on 5-sigma negative flux pixels.  That threshold can be changed with `--negflux-nsig` (specify a positive number of sigma, the threshold will be applied to negative flux).  Having written that, I could probably make this safer by adding an abs or something to prevent sign misunderstandings leading to masking positive data...

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -467,7 +467,7 @@ class DistTargetsDESI(DistTargets):
                             # mask significantly negative data (set ivar to 0)
                             sigma = self._my_data[toff].spectra[tspec_ivar[t]].flux * \
                                     np.sqrt(self._my_data[toff].spectra[tspec_ivar[t]].ivar)
-                            bignegflux = (sigma<-negflux_nsig)
+                            bignegflux = (sigma<-abs(negflux_nsig))
                             self._my_data[toff].spectra[tspec_ivar[t]].ivar[bignegflux] = 0.0
 
                             tspec_ivar[t] += 1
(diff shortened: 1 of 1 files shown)


sbailey:
@moustakas you've been the most extensive user of this branch.  Have you seen any evidence that masking significantly negative flux leads to a bias of false positive fits to unmasked positive flux, e.g. from sky mis-subtraction ringing?  Or have your tests been focused on other things and you couldn't say either way?

FYI, @rongpu is investigating improvements upstream, but we could still proceed with this PR as a smaller step anyway.

moustakas:
@sbailey, testing this branch has been pretty low priority, although I did run with it on about 70,000 objects as part of my most recent template testing and there were no major problems that I noticed. A more careful test would be to loop through spectra and just _count_ how many pixels would be masked, but I don't have any cycles to do this.

Anecdotally, this branch fixes this issue (and we get the correct redshift now!), so I'm happy to merge---
https://github.com/desihub/redrock/issues/281

For your consideration, I also think that this change should be the default (but perhaps in a separate PR, tied to Jura).

sbailey:
Merging now.  FYI, this change is the default at 5 sigma, i.e. the option can be used to change to to something other than 5 sigma, but if you don't specify anything you get a 5 sigma negative flux mask.

PR #278, opened 2024-02-01
Archetype failure due to key error in trans dict and failure in deg-leg 0 case

abhi0395:
This pull request solves two separate issues on the `main` branch.

1) For some targetids the archetype mode fails citing `keyerror` in the trans dictionary.
2) The main branch fails if we want to run the archetype method without Legendre polynomials.

Example run for case (1) on `main' branch: 

`rrdesi -i /global/cfs/cdirs/desi/spectro/redux/daily/tiles/cumulative/6786/20240117/coadd-4-6786-thru20240117.fits -o arch_test-39628278872412904.fits -d arch_39628278872412904.h5 --archetypes /global/homes/a/abhijeet/software/desisoft/new-archetypes/rrarchetype-galaxy.fits --targetids 39628278872412904`

fails citing following error:

```
MP calc_zchi2: Traceback (most recent call last):
MP calc_zchi2:   File "/global/homes/a/abhijeet/software/desisoft/redrock/py/redrock/zfind.py", line 90, in _mp_fitz
    zfit = fitz(chi2[i], t.template.redshifts, tg,
MP calc_zchi2:   File "/global/homes/a/abhijeet/software/desisoft/redrock/py/redrock/fitz.py", line 338, in fitz
    chi2min, coeff, fulltype = archetype.get_best_archetype(target,weights,flux,wflux,dwave,zbest, per_camera, n_nearest, trans=trans, use_gpu=use_gpu, prior=prior)
MP calc_zchi2:   File "/global/homes/a/abhijeet/software/desisoft/redrock/py/redrock/archetypes.py", line 321, in get_best_archetype
    if (trans[hs] is not None):
MP calc_zchi2: KeyError: -5909694164499214244
```

But when run on `archetype_leg0_and_trans_fix`, it succeeds.

Example run for case (2):

`rrdesi -i /global/cfs/cdirs/desi/spectro/redux/daily/tiles/cumulative/11472/20240116/coadd-0-11472-thru20240116.fits -o /global/cfs/cdirs/desi/users/abhijeet/test_archetype_runs/archetype_leg0_and_trans_fix/redrock_test-39628128854739056.fits -d /global/cfs/cdirs/desi/users/abhijeet/test_archetype_runs/archetype_leg0_and_trans_fix/redrock_39628128854739056.h5 --targetids 39628128854739056 --archetypes /global/homes/a/abhijeet/software/desisoft/new-archetypes/rrarchetype-galaxy.fits`

fails on `main' branch:

```
[...]

PR #287, opened 2024-03-12
Legendre nnls

craigwarner-ufastro:
Non-negative least squares (NNLS) is significantly faster than the general case bounded value least squares (BVLS).  For archetypes, BVLS had been used where the archetype term had a bounded solution from 0 to inf and all of the legendre term solutions were allowed to float.  By adding additional terms to our basis vector that are the inverse of each legendre term, we are able to use NNLS instead, e.g., we have negative copies, v3 = -v1 and v4 = -v2 so that in our solution we could have c1 > 0, c3 = 0 for a positive result and the opposite for a negative.

This also required a modification to the prior array which was added to the `M` vector.  The results using NNLS now agree to within np.allclose to the original BVLS solution, with a significant speed gain.  The `Finding best redshift` section takes 45s on 4 GPU + 4 CPU versus 99s using BVLS.  However this only applies to when using GPUs.

When using 64 CPUs, additional negative legendre terms + NNLS are 113s for this section versus 74s for BVLS.  This is because the computation time for various array operations such as dot products on the larger arrays with the added terms overwhelms the speed gains from NNLS vs BVLS and thus when using CPU mode, we do NOT use the NNLS method and maintain the original BVLS.

### DIFF
--- a/py/redrock/archetypes.py
+++ b/py/redrock/archetypes.py
@@ -299,9 +299,7 @@ class Archetype():
 
         obs_wave = np.concatenate([dwave[key] for key in new_keys])
         
-        nleg = legendre[list(legendre.keys())[0]].shape[0]
-        #zzchi2 = np.zeros(self._narch, dtype=np.float64)
-        #zzcoeff = np.zeros((self._narch,  1+ncam*(nleg)), dtype=np.float64)
+        #nleg = legendre[list(legendre.keys())[0]].shape[0]
         
         #TODO: return best fit model as well
         #zzmodel = np.zeros((self._narch, obs_wave.size), dtype=np.float64)
@@ -343,10 +341,11 @@ class Archetype():
                 tdata[hs] = binned[hs].transpose()[:,:,None]
             nbasis = tdata[hs].shape[2]
         if per_camera:
+            #Hard code NNLS as solve method for per_camera_coeff_with_least_square_batch
             if (use_gpu):
-                (zzchi2, zzcoeff) = per_camera_coeff_with_least_square_batch(target, tdata, gpuweights, gpuflux, gpuwflux, tot_nleg, self._narch, method=solve_method, n_nbh=1, prior=prior, use_gpu=use_gpu, ncam=ncam)
+                (zzchi2, zzcoeff) = per_camera_coeff_with_least_square_batch(target, tdata, gpuweights, gpuflux, gpuwflux, tot_nleg, self._narch, method='NNLS', n_nbh=1, prior=prior, use_gpu=use_gpu, ncam=ncam)
             else:
-                (zzchi2, zzcoeff) = per_camera_coeff_with_least_square_batch(target, tdata, weights, flux, wflux, tot_nleg, self._narch, method=solve_method, n_nbh=1, prior=prior, use_gpu=use_gpu, ncam=ncam)
+                (zzchi2, zzcoeff) = per_camera_coeff_with_least_square_batch(target, tdata, weights, flux, wflux, tot_nleg, self._narch, method='NNLS', n_nbh=1, prior=prior, use_gpu=use_gpu, ncam=ncam)
         else:
             if (use_gpu):
                 (zzchi2, zzcoeff) = calc_zchi2_batch(spectra, tdata, gpuweights, gpuflux, gpuwflux, self._narch, nbasis, use_gpu=use_gpu)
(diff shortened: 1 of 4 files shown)


craigwarner-ufastro:
Note: This also includes a bug fix related to #278

sbailey:
@craigwarner-ufastro thanks for this update and apologies for the late review.

First of all, a bug:
```
srun -n 4 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores rrdesi_mpi --gpu \
  -i $CFS/desi/spectro/redux/iron/tiles/cumulative/80605/20210205/coadd-6-80605-thru20210205.fits \
  --archetypes /global/cfs/cdirs/desi/users/abhijeet/new-archetypes/rrarchetype-galaxy.fits \
  -o $SCRATCH/redrock/redrock-legnnls.fits
...
--- Process 0 raised an exception ---
Proc 0: Traceback (most recent call last):
Proc 0:   File "/global/common/software/desi/users/sjbailey/redrock/py/redrock/external/desi.py", line 916, in rrdesi
    scandata, zfit = zfind(targets, dtemplates, mpprocs,
Proc 0:   File "/global/common/software/desi/users/sjbailey/redrock/py/redrock/zfind.py", line 602, in zfind
    if archetype._solver_method == 'bvls':
Proc 0: AttributeError: 'NoneType' object has no attribute '_solver_method'

MPICH Notice [Rank 0] [job id 24313752.5] [Fri Apr 12 09:37:16 2024] [nid002260] - Abort(0) (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 0) - process 0
```

Second, this update required a lot of delicate bookkeeping about the number of legendre coefficients, setting up the prior, etc. impacting multiple functions in multiple files.  I'm concerned about that from a maintenance perspective.  It feels like this could/should be an internal detail of the solver, not something that requires the "user" to setup that negative/positive legendre trick themselves and then unpack the results.  i.e. could you move all of this logic into `per_camera_coeff_with_least_square_batch` itself?  If it receives method='bvls' and use_gpu=True, it would automatically switch to the +/- legendre nnls method, including padding the arrays and expanding the prior as needed?  Can that be implemented with similar speed gains, while isolating all of the messy bookkeeping to a single location?

sbailey:
Thanks for the updates, @craigwarner-ufastro .  Please resolve the merge conflicts that have arisen here due to other merge-conflicts.  If they aren't trivial to resolve in-place, please *don't* merge main into this branch, which would make the changes in this PR harder to evaluate mixed in with all the other changes.  If needed, ok to make a new branch from this branch and then rebase that off of main and start a new PR.  Ping me on Slack if that operation description is confusing.

Back to the updates: I'd like to take this one step further to avoid legendre-specific manipulations to tdata prior to calling the solver, e.g. in `archetypes.nearest_neighbor_model`:
```python
tdata[hs] = np.append(tdata[hs], legendre[hs].transpose()[None,:,:], axis=2)
                if per_camera and self._solver_method == 'bvls' and use_gpu:
                    #Using BVLS and GPU - append negative copies of legendre terms as well to use NNLS with additional bases to approximate BVLS
                    tdata[hs] = np.append(tdata[hs], -1*legendre[hs].transpose()[None,:,:], axis=2)
```
and in `zfind.zfind`:
```python
        tot_deg_legendre = deg_legendre
        for key in archetypes:
            if archetypes[key]._solver_method == 'bvls':
                tot_deg_legendre = deg_legendre*2
...
        maxcoeff = max(np.max([t.template.nbasis for t in templates]), ncamera*(tot_deg_legendre)+n_nearest)
```
in the second case, I'm not even sure that is correct because we are reserving `deg_legendre*2` coefficients but those extra 2x coefficients are for the BVLS -> NNLS trick and shouldn't be needed for the output arrays.

Suggestion: rather than manipulating `tdata` (template data) to add legendre terms and then calling `per_camera_coeff_with_least_square_batch(... tdata, nleg=...)`, could you pass `tdata` and the normal `legendre` array, and let `per_camera_coeff_with_least_square_batch` derive `nleg` from the shape of `legendre` and whether it needs to do the BLVS -> NNLS trick and tdata manipulation internally based upon `method` and `use_gpu`?

i.e. from a user-perspective, turn this into "if you want to include legendre terms, pass in the legendre template arrays to use" and all other details of how those are combined with `tdata` etc. become internal to `per_camera_coeff_with_least_square_batch`.  Thoughts?

PR #289, opened 2024-03-29
add select_targetids optional input to results.read_zscan

moustakas:
@sbailey this should be a non-controversial change, but it's needed as part of the Y3 template work.

### DIFF
--- a/py/redrock/results.py
+++ b/py/redrock/results.py
@@ -108,9 +108,6 @@ def read_zscan(filename, select_targetids=None):
                 - zwarn: 0=good, non-0 is a warning flag
 
     """
-    def encode_column(c):
-        return c.astype((str, c.dtype.itemsize))
-    
     import h5py
     # zbest = Table.read(filename, format='hdf5', path='zbest')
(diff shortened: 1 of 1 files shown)


py/redrock/results.py
```diff
@@ -104,14 +108,22 @@ def read_zscan(filename):
                 - zwarn: 0=good, non-0 is a warning flag
 
     """
+    def encode_column(c):
+        return c.astype((str, c.dtype.itemsize))
```
sbailey, line 112: Is this a debugging leftover?  Why is this function here, replacing the one imported via `from .utils import encode_column`.  If we specifically need it here, please also copy the docstring.

sbailey (commented):
Looks good, but I put a question inline about a potentially leftover debugging change.

py/redrock/results.py
moustakas: I honestly have no idea why I put this here (which was some time ago...). I'll remove it for this PR but if I figure out why I tripped on it in the first place I'll follow up!

Issue #291, opened 2024-04-18
inconsistent legendre wavelength basis


sbailey:
The conversion of wavelength to reduced wavelength for Legendre polynomials currently happens in 3 different places in the Redrock code:

```
py/redrock/targets.py:186:     self._legendre = { hs:np.array([scipy.special.legendre(i)( (w-wmin)/(wmax-wmin)*2.) for i in range(nleg)]) for hs, w in dwave.items() }
py/redrock/archetypes.py:161:  legendre = np.array([scipy.special.legendre(i)( (wave-wave_min)/(wave_max-wave_min)*2.-1. ) for i in range(deg_legendre)])
py/redrock/fitz.py:115:        legendre = { hs:np.array([scipy.special.legendre(i)( (w-wmin)/(wmax-wmin)*2.) for i in range(nleg)]) for hs, w in dwave.items() }
```

The definition in archetypes.py line 161 is correct to transform it to a [-1,1] range:

(wave - wave_min) / (wave_max - wave_min) * 2 - 1

which is the same as

((wave - wave_min) / (wave_max - wave_min)) * 2 - 1

where I added extra parentheses to emphasize that the division happens before the *2 multiplication, i.e. the 2 is multiplied by the numerator, not the denominator.

The other two definitions are missing the "-1" term and thus have a range of [0,2].

Clean this up to be both correct and consistent.  I suggest that we add a utility function to do this transformation in a single place, e.g.

```python
def reduced_wavelength(wave):
    wave = np.asarray(wave)
    wavemax = wave.max()
    wavemin = wave.min()
    return 2*(wave - wavemax) / (wavemax - wavemin) - 1
```

moustakas:
Done in #299.

Issue #274, opened 2024-01-26
COEFF: track PCA/NMF vs. Archetypes vs. Legendre separately


sbailey:
Currently the Redrock output COEFF column can mean different things:
  * in default mode it is the PCA/NMF template coefficients
  * in archetype mode, it is the archetype coefficients, and the PCA/NMF coefficients are lost
  * in archetype+legendre mode some coefficients are archetype coeffs and others are legendre coeffs
    * I'm not even sure if you can tell from the output file whether it used 1+3*n archetypes, or 1 archetype with 3 cameras * n legendre coefficients.

I suggest that we track these separately with something like:
  * COEFF is always coefficients from original PCA/NMF scan
  * ARCHCOEFF (or some similar name) are optional Archetype coefficients
  * LEGCOEFF are optional Legendre coefficients
    * TBD: or maybe LEGCOEFF if those apply to COEFF, and ARCHLEGCOEFF if they apply to Archetypes?
    * TBD: for LEGCOEFF[i] as the coefficients for target i:
      * if this is 1D[n], it means they are n polynomials cross-camera (like the templates); 
      * if this is 2D[m,n], it means they are for m cameras and n polynomials.

@abhi0395 let's discuss and refine.  Comments / ideas from others are welcome too.

PR #293, opened 2024-04-19
refactor template/archetype model eval

sbailey:
@abhi0395 this is an update to your best_fit_model branch (PR #283) to address some of my maintainability concerns.  I've made enough changes that I'm doing this as a PR to your branch rather than updating your branch directly.  Although this version is quite a bit different from your original branch, I learned a *lot* from your work about the various corner cases with MPI vs. multiprocessing, wavehashes out of order compared to camera bands, etc. so thanks for that initial solution which guided me for what to watch out for.

In the end the output files are the same except
* I renamed "COEFFTYPE" to "FITMETHOD" in anticipation of issue #274 splitting archetype coefficients into a separate column from template coefficients, and I also promoted that into being written out as part of the regular redrock REDSHIFTS table, not just in the model file.
* for archetypes with nearest neighbors, I joined the subtypes with a semicolon instead of an underscore so that e.g. SUBTYPE='ELG_23;ELG_40' instead of SUBTYPE='ELG_23_ELG_40'.  That makes it easier to split the SUBTYPE string into the individual archetype subtypes.

**Big Caveat**: the evaluated Archetype+legendre models are not correct, which I think is due to issue #291 with the inconsistent definitions of legendre basis wavelengths and whether they span the entire brz wavelength range or only the individual cameras.  Let's get that sorted out separately, and then revisit this model evaluation to make sure it uses the same pieces.  We don't need this for Jura, but we will need it for Archetype runs post-Jura.

## More details

The goal here is to have `Template.eval()` and `Archetypes.eval()` "own" the concept of how to evaluate a model given coefficients and have `rrdesi -> DistTargets.eval_models -> Target.eval_model` use those.  This simplifies the bookkeeping to avoid things like
* wavedict, wave_dict, and dwave all being dictionaries of wavelength grids differing only by their keys
* `Archtype.get_best_archetype_model` having to know about templates just in case it is asked to evaluate something that isn't an archtype.
* New functions `get_best_model_spectra`, `eval_model_for_one_spectra `, and `eval_tdata` having significant conceptual overlap with previously existing `eval` functions, but not actually using those.  Instead of pulled pieces of those into the the other eval functions.  

I also updated some of the underlying messiness that made your original branch hard to implement:
* wavehash vs. camera bands: For DESI, all spectra of a given band have the same wavelengths so DistTargetsDESI can just use the band as the wavehash.
* wavehash out of order compared to camera bands (this was due to a `set` operation changing the order).

Please take a look at this, and then let's discuss any items that you disagree with and/or have a better idea for how to implement.  Thanks.

### DIFF
--- a/py/redrock/archetypes.py
+++ b/py/redrock/archetypes.py
@@ -20,8 +20,6 @@ from .igm import transmission_Lyman
 
 from .zscan import per_camera_coeff_with_least_square_batch
 
-from .templates import eval_model_for_one_spectra
-
 class Archetype():
     """Class to store all different archetypes from the same spectype.
 
@@ -66,6 +64,8 @@ class Archetype():
 
         # TODO: Allow Archetype files to specify their IGM model
         self.igm_model = 'Inoue14'
+
+        self.method = 'ARCH'  # for API symmetry with Template.method
 
         return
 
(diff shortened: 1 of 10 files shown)


abhi0395 (approved):
This looks good to me. Thanks for cleaning out the messiness of my best_fit_model branch. I have also opened the [PR#299](https://github.com/desihub/redrock/pull/299), which fixes the inconsistent definition of legendre wavelength basis. 

I think, merging this branch with that should fix legendre wavelength basis definition here as well.

Issue #248, opened 2023-05-23
Are penalties actually applied at the final sorting?


Waelthus:
I looked at the zscan and fitz routines again and found a potential bug regarding final sorting:

- in the initial coarse redshift fitting, penalties for forbidden lines are applied (in the zscan.calc_zchi2 function: https://github.com/desihub/redrock/blob/2483c484cbccd6c785126c2d045e02deccf1ffe7/py/redrock/zscan.py#L746
- afterwards fits are performed within the fitz.fitz function which first gets the penalized chi2 values, searches for minima and then refits on a finer wavelength grid 
- this refit is done using zscan.calc_zchi2_batch https://github.com/desihub/redrock/blob/2483c484cbccd6c785126c2d045e02deccf1ffe7/py/redrock/fitz.py#L210 which does not report a penalty and does not use calc_zchi2 internally, so won't compute a penalty https://github.com/desihub/redrock/blob/2483c484cbccd6c785126c2d045e02deccf1ffe7/py/redrock/zscan.py#L657
- those chi2 values are then fit with a parabola https://github.com/desihub/redrock/blob/2483c484cbccd6c785126c2d045e02deccf1ffe7/py/redrock/fitz.py#L217 and the outputs of that fit are then used for the final sorting https://github.com/desihub/redrock/blob/2483c484cbccd6c785126c2d045e02deccf1ffe7/py/redrock/fitz.py#L289

So it looks like we're currently only applying the penalties on coarse fits and both the final reported chi2 and the final sort order are not including a penalty term.
How much this actually affects final redshifts remains to be seen.

moustakas:
Possibly addressed in https://github.com/desihub/redrock/pull/296.

sbailey:
Fixed by PR #296.  Closing.

PR #296, opened 2024-04-23
include negative [OII] flux penalty in final minimization

moustakas:
This PR is a simple attempt to address #248. I haven't tested it extensively, but trying to get it up for review sooner rather than later.

Note that the ad hoc "penalize negative [OII] flux" should go away (or be turned off) once we move to NMF templates.

### DIFF
--- a/py/redrock/fitz.py
+++ b/py/redrock/fitz.py
@@ -16,6 +16,7 @@ from . import constants
 from .rebin import rebin_template
 
 from .zscan import calc_zchi2_one, spectral_data, calc_zchi2_batch
+from .zscan import calc_negOII_penalty
 
 from .zwarning import ZWarningMask as ZW
 
@@ -259,11 +260,8 @@ def fitz(zchi2, redshifts, target, template, nminima=3, archetype=None, use_gpu=
                                                  use_gpu=use_gpu)
 
         #- Penalize chi2 for negative [OII] flux; ad-hoc
-        zzchi2penalty = zzchi2 * 0.
-        if template.template_type == 'GALAXY':
-            OIIflux = np.sum(zzcoeff @ OIItemplate.T, axis=1)
-            zzchi2penalty[OIIflux < 0] = -OIIflux[OIIflux < 0]
-        zzchi2 += zzchi2penalty
+        if hasattr(template, 'OIItemplate'):
+            zzchi2 += calc_negOII_penalty(template.OIItemplate, zzcoeff)
 
         #- fit parabola to 3 points around minimum
         i = min(max(np.argmin(zzchi2),1), len(zz)-2)
--- a/py/redrock/fitz.py
+++ b/py/redrock/fitz.py
@@ -136,7 +136,8 @@ def prior_on_coeffs(n_nbh, deg_legendre, sigma, ncamera):
     return prior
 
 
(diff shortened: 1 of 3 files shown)


sbailey:
@moustakas thanks for pushing this forward.  I took your implementation one step further to combine the negative [OII] penalty logic into a single function called from both places rather than having various wavelength ranges, matrix algebra, sums, etc. replicated in two different places with slightly different variable names.

I ran this version and main on the same input file, and verified that the coarse scan zchi2 and penalty vs. z remain unchanged compared to main, but the local minimal zfit results do change for some targets due to the newly added penalty term in the final fits.  7/500 were more than 1e-3, and only one of those originally had ZWARN=0 (and a DELTACHI2=18).

> Note that the ad hoc "penalize negative [OII] flux" should go away (or be turned off) once we move to NMF templates.

Agreed.  I suppose I could have put a check in that is method=NMF, don't even bother doing the calculation because by construction it can't be negative, but that isn't currently implemented.

py/redrock/fitz.py
```diff
@@ -252,6 +259,10 @@ def fitz(zchi2, redshifts, target, template, nminima=3, archetype=None, use_gpu=
                                                  solve_matrices_algorithm=template.solve_matrices_algorithm,
                                                  use_gpu=use_gpu)
 
+        #- Penalize chi2 for negative [OII] flux; ad-hoc
+        if hasattr(template, 'OIItemplate'):
+            zzchi2 += calc_negOII_penalty(template.OIItemplate, zzcoeff)
+
```
moustakas, line 265: I don't think this is right. `OIItemplate` is defined above as a numpy array but not as an attribute of the `template` object. Or am I missing something?

moustakas:
@sbailey please take a look at my changes. I put in the restriction that the [OII] penalty should not be applied to NMF or NNLS fitting or templates.

py/redrock/fitz.py
moustakas: Please ignore this comment; I saw after I wrote this that `OIItemplate` was being set in `redrock.templates` (and that the code above was vestigial and not needed).

sbailey: Thanks for catching that and removing it..  I had meant to remove that and forgot.  Moving it into the Templates class was so that there would be a single place defining the wavelength range and the concept of the OIItemplate.

py/redrock/templates.py
```diff
@@ -158,6 +158,13 @@ def __init__(self, filename=None, spectype=None, redshifts=None,
             raise ValueError(f'Template method {self._method} unrecognized; '
                               f'should be one of {valid_template_methods}')
 
+        #- Special case for GALAXY templates: cache subset of template that
+        #- covers [OII] but only for GALAXY templates and fitting methods which
+        #- can include negative templates or coefficients.
+        if self._rrtype == 'GALAXY' and not self._method in ("NMF", "NNLS"):
```
sbailey, line 164: Good update.  I'd even be fine with more restrictive
```python
if self._rrtype == 'GALAXY' and self._method == 'PCA':
    ...
```
Thoughts?

moustakas:
> Good update. I'd even be fine with more restrictive
>     `if self._rrtype == 'GALAXY' and self._method == 'PCA':`

I agree. I made the change; this should be good to merge, which should close #248.

PR #295, opened 2024-04-22
add TEMFILnn keywords with full template filenames

sbailey:
@moustakas this is the followup PR to #290 to save the template filenames in the output headers as TEMFILnn keywods to support non-standard template names.  This is supported by `redrock.templates.load_templates_from_header` to read them back in.

Example:
```
echo $RR_TEMPLATE_DIR
cat $RR_TEMPLATE_DIR/templates-test.txt
rrdesi -n 16 \
  -i $CFS/desi/spectro/redux/iron/tiles/cumulative/80605/20210205/coadd-6-80605-thru20210205.fits \
  -t $RR_TEMPLATE_DIR/templates-test.txt \
  -o $SCRATCH/redrock/redrock-test-keywords.fits
fitsheader -e 0 $SCRATCH/redrock/redrock-test-keywords.fits
```

Results:
```
/global/common/software/desi/users/sjbailey/redrock-templates
rrtemplate-GALAXY-None-v2.6.fits
rrtemplate-QSO-SuperDuperWow.fits
...
TEMNAM00= 'GALAXY  '                                                            
TEMVER00= '2.6     '                                                            
TEMFIL00= 'rrtemplate-GALAXY-None-v2.6.fits'                                    
TEMNAM01= 'QSO:::LOZ'                                                           
TEMVER01= '1.0     '                                                            
TEMFIL01= 'rrtemplate-QSO-SuperDuperWow.fits'                                   
...
DEPNAM19= 'RR_TEMPLATE_DIR'                                                     
DEPVER19= '/global/common/software/desi/users/sjbailey/redrock-templates'       
```

Loading those templates using the header:
```python
import fitsio
from redrock.templates import load_templates_from_header
hdr = fitsio.read_header('/pscratch/sd/s/sjbailey/redrock/redrock-test-keywords.fits')
templates = load_templates_from_header(hdr)
```
Results:
```
Reading templates from ['/global/common/software/desi/users/sjbailey/redrock-templates/rrtemplate-GALAXY-None-v2.6.fits', '/global/common/software/desi/users/sjbailey/redrock-templates/rrtemplate-QSO-SuperDuperWow.fits']
INFO: rrtemplate-GALAXY-None-v2.6.fits GALAXY  PCA IGM=Calura12 z=-0.005-1.6997
INFO: rrtemplate-QSO-SuperDuperWow.fits QSO LOZ PCA IGM=Calura12 z=0.05-1.5983
```

[...]

PR #297, opened 2024-04-24
fix read_zscan; add make_fulltype, parse_fulltype

sbailey:
@moustakas this PR fixes the bug you discovered in PR #295, where read_zscan was returning results missing the QSO:::LOZ zfit table.  It turned out to be a spectype vs. fulltype=spectype:::subtype confusion.  I took the opportunity to add `fulltype = make_fulltype(spectype, subtype)` and `spectype, subtype = parse_fulltype(fulltype)` helper functions instead of fragile parsing in multiple locations.

Your example from PR #295 should now work, using the details file that you generated before.

Side note: the details file keeps the best 3 fits per SPECTYPE, but not per (SPECTYPE, SUBTYPE).  So if you are fitting both QSO:::LOZ and QSO:::HIZ and all of the best fits are QSO:::HIZ, then the details file doesn't have zfit info for the QSO:::LOZ minima and the QSO:::LOZ zfit table will still be length 0, while the QSO:::HIZ table will be length 3.  In the example case you ran, the best fits were QSO:::LOZ and they should have been in the table, but weren't due to the spectype vs. fulltype bug fixed in this PR.


moustakas (approved):
Looks great, thank you.

PR #299, opened 2024-04-24
fixing inconsistent legendre wavelength basis (#291)

abhi0395:
Thanks @sbailey for pointing me to the issue of inconsistent legendre wavelength basis definition in code. The issue is detailed in [issue#291](https://github.com/desihub/redrock/issues/291) have now defined a separate utility function `reduced_wavelength()` in `utils.py`, which is called whenever we want to define legendre wavelength basis. It makes the code cleaner and consistent in archetype mode. This consistent definition should also be included in [PR#293](https://github.com/desihub/redrock/pull/293) and [PR#283](https://github.com/desihub/redrock/pull/283).

It also fixes a minor case, when an end user does not want to add a prior in the archetype mode. I have implemented a minor logic, that if `--archetype-legendre-prior` flag is <=0, then no prior will be added in archetype mode.

### DIFF
--- a/py/redrock/archetypes.py
+++ b/py/redrock/archetypes.py
@@ -157,9 +157,6 @@ class Archetype():
         deg_legendre = (coeff!=0.).size-1
         index = np.arange(self._narch)[self._subtype==subtype][0]
 
-        #w = np.concatenate([ w for w in dwave.values() ])
-        #wave_min = w.min()
-        #wave_max = w.max()
         legendre = np.array([scipy.special.legendre(i)(reduced_wavelength(w)) for i in range(deg_legendre)])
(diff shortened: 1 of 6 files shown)


sbailey:
Thanks.  For the record, this PR also updates the wavelength -> [-1,1] mapping to be per-camera instead of using the full wavelength basis across all cameras.  I think that is what we want so that the legendre polynomials are orthogonal per camera.

I also added unit tests on `reduced_wavelength` which caught a `wavemax` vs. `wavemin` bug which I corrected.

abhi0395:
Thank you so much. Luckily, I hadn't run anything big yet.

PR #310, opened 2024-06-03
point2point gather of results

sbailey:
This PR fixes a memory/objsize problem in Jura.  Healpix with an especially large number of targets were failing on the final gather:
```python
results = targets.comm.gather(results, root=0)
...
    results = targets.comm.gather(results, root=0)
Proc 0:   File "mpi4py/MPI/Comm.pyx", line 1578, in mpi4py.MPI.Comm.gather
Proc 0:   File "mpi4py/MPI/msgpickle.pxi", line 772, in mpi4py.MPI.PyMPI_gather
Proc 0:   File "mpi4py/MPI/msgpickle.pxi", line 777, in mpi4py.MPI.PyMPI_gather
Proc 0:   File "mpi4py/MPI/msgpickle.pxi", line 190, in mpi4py.MPI.pickle_allocv
Proc 0:   File "mpi4py/MPI/msgpickle.pxi", line 181, in mpi4py.MPI.pickle_alloc
Proc 0: SystemError: Negative size passed to PyBytes_FromStringAndSize
```
This would occur even if the individual ranks had fairly small inputs (in particular below 2GB).

This PR works around this by doing multiple point-to-point send/recv transfers instead of a gather.  Admittedly, this might just be buying us time as healpix continue to grow, but it works on special-other healpix 27239 with 12590 targets, the largest case in Jura that has made it to the Redrock step.  If needed, the next step in bookkeeping would be to split the `results` dict into multiple smaller dicts and transfer them, but I haven't tackled that yet.

In addition to testing several very large healpix, I did a direct comparison of main-dark healpix 0 with main and this PR branch in /pscratch/sd/s/sjbailey/redrock/redrock-0-{main,pr}.fits, which passes fitsdiff (data are same, timestamps and dependency keywords diff).  They also take the same amount of time so that multi send/recv isn't hurting performance.

### DIFF
--- a/py/redrock/zfind.py
+++ b/py/redrock/zfind.py
@@ -403,7 +403,17 @@ def zfind(targets, templates, mp_procs=1, nminima=3, archetypes=None, priors=Non
     allzfit = None
 
     if targets.comm is not None:
-        results = targets.comm.gather(results, root=0)
+        # gather can fail with especially large inputs, so use point-to-point
+        ### results = targets.comm.gather(results, root=0)
+        if targets.comm.rank == 0:
+            results = [results,]
+            for other_rank in range(1,targets.comm.size):
+                other_results = targets.comm.recv(source=other_rank, tag=42)
+                results.append(other_results)
+        else:
+            targets.comm.send(results, dest=0, tag=42)
+
+        targets.comm.barrier()
     else:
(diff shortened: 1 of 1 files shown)


dmargala (approved):
LGTM!

PR #311, opened 2024-06-04
more MPI memory management on large inputs

sbailey:
This PR makes 3 updates to help with MPI memory management on especially large inputs like jura special-other healpix 27257 with 33749 targets (!):
  * when broadcasting the large RESOLUTION HDUs, break into chunks to avoid 2 GB per-bcast limit
  * when collecting results, break into chunks to avoid 2 GB per-send limit
    * this is a followup to PR #310 which worked up to ~20k targets, but not ~30k
  * if the input communicator is larger than needed, split it into a sub communicator with just the ranks that will be use to avoid unnecessary bcasts to unused ranks

I verified on main/dark healpix 0 that the outputs are the same as main, whether the chunks are big or small.  A final test on special-other healpix 27257 is still running; I will confirm that succeeded before merging, but I wanted to get this PR in while that was running.

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -570,7 +570,34 @@ class DistTargetsDESI(DistTargets):
                     hdata = hdus[extname].data[rows]
 
                 if comm is not None:
-                    hdata = comm.bcast(hdata, root=0)
+                    #- resolution data for dense healpix can exceed the
+                    #- mpi4py 2GB limit, so distribute in batches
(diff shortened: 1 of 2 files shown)


sbailey:
Test job succeeded on 33749 targets in 25 minutes.  After discussion with @dmargala I removed the unnecessary send/recv tag logic, since the blocking operations enforce the order anyway.  Will merge and retag for use with Jura.  This is backwards compatible with output of previous version.

PR #313, opened 2024-07-18
Clean-up environment variables moved to desimodules

weaverba137:
This PR cleans up environment variables that are now defined in desimodules as of desihub/desimodules#52.


weaverba137:
The Python 3.10 tests are pulling in Numpy 2.0, so that will need to be pinned.

sbailey (approved):

PR #320, opened 2024-10-29
support rrdesi --templates (list of templates)

sbailey:
This PR adds support for `rrdesi --templates ...` to accept a list of templates, instead of previously only accepting a single template or a path to a text file with a list of templates.  The intended use case is for the DESI QuasarNet afterburner which filters the template files down to just QSO templates and then wants to run redrock with just those.  It could also be useful in the future for e.g. DESI-2 running with LAE/LBG templates vs. standard galaxy+QSO+star templates, etc.

Alternatively, we could have added a templates-default-qso.txt file to the redrock-templates repo have have the QN afterburner use that.  That would also work (and is still compatible with this PR), but it becomes a bit more of a maintenance headache when making tags and also needing to make templates-X.Y-qso.txt, and when updating QSO templates and needing to update multiple templates-*.txt files.

I updated the test to include this new functionality, and also tested with
```
srun -n 4 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores rrdesi_mpi --gpu -i coadd-0-100-thru20210505.fits -o $SCRATCH/temp/redrock.fits --templates $RR_TEMPLATE_DIR/rrtemplate-QSO-LOZ-v1.0.fits $RR_TEMPLATE_DIR/rrtemplate-QSO-HIZ-v1.1.fits
```
with output in
```
/pscratch/sd/s/sjbailey/temp/redrock.fits
```
confirming that it only used QSO templates (which was also clear from the logs):
```
[login33 20210505] fitsheader -e 0 /pscratch/sd/s/sjbailey/temp/redrock.fits | grep ^TEM
TEMNAM00= 'QSO:::LOZ'                                                           
TEMVER00= '1.0     '                                                            
TEMFIL00= 'rrtemplate-QSO-LOZ-v1.0.fits'                                        
TEMNAM01= 'QSO:::HIZ'                                                           
TEMVER01= '1.1     '                                                            
TEMFIL01= 'rrtemplate-QSO-HIZ-v1.1.fits'                                        
```

Changes still need to be made on the desispec side to use this, but this adds the feature.

### DIFF
--- a/py/redrock/external/desi.py
+++ b/py/redrock/external/desi.py
@@ -660,7 +660,7 @@ def rrdesi(options=None, comm=None):
     parser = argparse.ArgumentParser(description="Estimate redshifts from"
         " DESI target spectra.")
 
-    parser.add_argument("-t", "--templates", type=str, default=None,
+    parser.add_argument("-t", "--templates", type=str, nargs='+', default=None,
         required=False, help="template file or directory")
 
     parser.add_argument("--archetypes", type=str, default=None,
--- a/py/redrock/fitz.py
+++ b/py/redrock/fitz.py
@@ -347,7 +347,14 @@ def fitz(zchi2, redshifts, target, template, nminima=3, archetype=None, use_gpu=
     ii = np.argsort([tmp['chi2'] for tmp in results])
     results = [results[i] for i in ii]
 
-    assert len(results) > 0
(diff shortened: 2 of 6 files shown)


abrodze (approved):
All changes look ok to me. I was able to run the branch with no issue

PR #347, opened 2026-02-20
Downgrade rr columns to lower precision

rongpu:
This PR makes the following data type changes for the DESI redrock catalog (redrock-*.fits) to reduce the file size. Also see https://github.com/desihub/desispec/issues/2506.

Column | Original Type → New Type
-- | -- 
ZERR | float64 → float32 |  
ZWARN | int64 → int32 |
SUBTYPE | 20 characters → 3 characters (or maximum length; see below) |  
NPIXELS | int64 → int32 |
NCOEFF | int64 → int16 |
COEFF | float64 → float32 |  
CHI2 | float64 → float32 |  
DELTACHI2 | float64 → float32 |  

For SUBTYPE, if any element is more than 3 character long, the column adopts the maximum length; otherwise it is 3 characters in size.


rongpu:
I have tested this by running redrock on a DESI coadd file, and the resulting catalog is as expected.

